Issue #389, opened 2015-01-07
create an enum of mongoose collection name strings to avoid hardcoding


tongxiang:
creating an enum of the collection names so that we don't have to hardcode a string each time we use `app.getConfig()`. Should save us from the occasional typo error or collection name changes. So end goal could look something this: app.getConfig(Config.COMPETITIVE_STORIES, id)

tongxiang:
Also in this issue--change the names of models to more accurately reflect if they're singular or plural. 

`startCampaignTransitions` --> `startCampaignTransition` 

`competitive_stories` --> `competitive_story`

PR #395, opened 2015-01-08
Scripts to pull config settings from production

### LINKED ISSUE #394: Create dev scripts for config deploy and pull
These tools are intended to be used in the interim while a frontend is being created.

- A script for team members to deploy .json config changes to our production database and restart the servers
- A script for devs to pull the configs from production and import them onto local dev environment

jonuy:
#### What's this PR do?
Provides a way to update the configs on production and to update your local dev box with changes on production. So our workflow after this would look like:

###### Set config variables
1. Create a `.deploy_config.conf` file and set the appropriate variables.

###### Deploy config changes to production
1. Modify relevant json files
2. `$ sh deploy_config.sh`

###### Pull config changes from production
1. `$ sh pull_config.sh`

Testing the deploy would actually deploy to production and restart the server, so maybe let's not do that one until we have real change we want to deploy.

But go ahead and run `sh pull_config.sh` to pull the contents of production onto your local config db.

Closes #394 

#### Questions
- Names of files? `deploy_config.conf` is really just a readme to tell you to make `.deploy_config.conf`. Then that `.deploy_config.conf` actually also gets used in the `pull_config.sh`. I'm just unsure if the names used here make sense.
- I removed database seeding from the wercker deploy step. Since we now will have this `deploy_config.sh` that can be run whenever, it seemed that having that as the one and only place where configs get updated would make things clearer(?). Thoughts?


deploy_config.sh
```diff
@@ -0,0 +1,77 @@
...
+
+    # Restart the app so the new configs can be loaded
+    jitsu apps restart $SMS_CONFIG_JITSU_APP_NAME
+
+    echo "\n\nApp restarted with new configs! Done!\n\n"
+  else
+    echo "Missing environment variables. Unable to push config to production. See deploy_config.conf for details."
+  fi
+else
+  echo "whew. ok then."
```
tongxiang: :smile_cat:

deploy_config.sh
```diff
@@ -0,0 +1,77 @@
...
+
+#
+# Push config settings based on local config .json files to the production database.
+#
+
+# Prompt user to make sure they don't do this by accident
+echo
+echo "***********************"
+echo "**    DANGER ZONE    **"
+echo "***********************"
```
tongxiang: image

tongxiang:
Would now be a good time to think about ways we can minimize changing things in multiple places when we need to add a new config collection type? (i.e., collaborative stories.) 

We talked before about some sort of enum (#389) to avoid hard-coding collection names when using app.getConfig(). Could there be a similar principle here, with the Mongo shell commands?

seed.sh
```diff
@@ -1,21 +1,7 @@
 #!/bin/bash
```
tongxiang: Within this file, do we even want to keep lines 20-30? (For some reason, Github won't let me comment down there.) 

Just in terms of avoiding confusion, considering that our workflow now emphasizes production's config as the source of truth, rather than the JSON files.

tongxiang:
So I understand that the point of running `deploy_config.sh` is to avoid having to re-deploy the app itself. 

1. If this step explicitly does **not** deploy our responder, could we name the file without the word "deploy"? Like `push_to_prod_config.sh`? Or `push_config`?
2. Also, given that our current stated workflow involves modifying JSON files, should we also push the modified JSON to master/production as well, to avoid merge conflicts in the future? 
3. If we do want to push the JSON to master/production, does this mean that we don't need to restart the app through jitsu?

pull_config.sh
```diff
@@ -0,0 +1,42 @@
...
+#
+
+# Prompt the user before running
+echo
+echo "*******************"
+echo "**    WARNING    **"
+echo "*******************"
+echo
+
+read -p "You're about to update your local config database with the one on production. You sure about this? [y/n] " -r
```
tongxiang: Should we specify here that existing local documents sharing the same `_id` value with production documents will **not** be overwritten? 

>If you restore to an existing database, mongorestore will only insert into the existing database, and does not perform >updates of any kind. If existing documents have the same value _id field in the target database and collection, >mongorestore will not overwrite those documents.

http://docs.mongodb.org/manual/reference/program/mongorestore/

deploy_config.sh
```diff
@@ -0,0 +1,77 @@
...
+        echo "jitsu not installed, trying to install it through npm"
+
+        if ! type npm &> /dev/null ; then
+            echo "npm not found, make sure you have npm or jitsu installed"
+        else
+            echo "npm is available"
+            echo "npm version: $(npm --version)"
+
+            echo "installing jitsu"
+            sudo npm install -g --silent jitsu
```
tongxiang: I was told before to avoid doing this: ["sudo npm install -g is an antipattern"](https://jesse.sh/sudo-npm-install/)

seed.sh
jonuy: > We talked before about some sort of enum (#389) to avoid hard-coding collection names when using app.getConfig(). Could there be a similar principle here, with the Mongo shell commands?

Maybe... I still feel ok about how this currently is because the extra level abstraction doesn't seem like it would help too much.

> Within this file, do we even want to keep lines 20-30?

Again, maybe... but this might still be needed for updating our local config collections.

pull_config.sh
jonuy: 1. File change to `push_config.sh`: cool
2. Does it seem ok to leave this responsibility to the person making the changes for now? Eventually, we'll be working towards removing the config/*.json files from the repo entirely.
3. The jitsu restart - we do still need to do it for it take the config changes, right? A good goal for later that we talked about would be a way to push changes and not have to restart the app. Probably outside the scope of this PR though.
4. And specifying that mongorestore won't overwrite documents (I think mongoimport does the same): cool

deploy_config.sh
jonuy: Gotcha. I'm ok just erroring out here instead and telling devs to install npm themselves. I shall go do that.

jonuy:
@tongxiang so this PR has morphed a bit from where it started out.

Things to look at now:
1. config/*.json files have been removed from the rep
2. `pull_config.sh` will be used by devs to update their local db from production if ever we need to
3. `seed.sh` has been updated to only be used by wercker

jonuy:
This should also close #397

tongxiang:
:+1:

.gitignore
```diff
@@ -9,3 +9,7 @@ newrelic_agent.log
 # test coverage
 .coveralls.yml
 coverage
+
+# config
+**/config/*.json
```
tongxiang: nice!

PR #399, opened 2015-01-13
Report Back Config Update

### LINKED ISSUE #396: Update report back configs with current campaigns and convert to db collection
- Add configs from current campaigns on the Drupal site to `reportback-config.json`
- Add the `reportback-config.json` to the process being put in place to update these configs on production

jonuy:
#### What's this PR do?
Updates the reportback code to use configs set in the database instead of in the static json file.

###### reportbackConfigModel.js
Added this model for defining reportback configs in the db.

###### smsConfigsLoader.js
The new model got added to the config array. Also updated `app.getConfig` to take a third parameter where you could specify they key you want to match against. You'll see that happening in reportback/index.js with:
```
campaignConfig = app.getConfig('reportbacks', campaign, 'endpoint');
```

Ran through a test on Postman with the `test` endpoint after importing configs to my local database. Also made sure that `npm test` worked.

Closes #396


tongxiang:
:+1: Sweet! Will add the reportback config info to the wiki documentation.

tongxiang:
Oh one minor thing--for consistency's sake, should we name the reportback model "reportback" instead of "reportbacks"? (Like before, as with `donorschoose`, we can set the third param in the model creation function to explicitly specify the name of the collection, if the pluralization is an issue.)

PR #404, opened 2015-02-03
Report Back Retry

### LINKED ISSUE #402: Output error log & retry report back submission when content API returns FALSE
There's a bug where report backs wouldn't successfully submit. They'd manifest themselves in logs with lines like:
```
Successfully submitted report back. rbid:
```
and rbid comes back empty/false.

Given the checks in place right now, the app assumes the API succeeded. We instead now want to both log that an error in submission has occurred and retry the POST request if possible.

jonuy:
#### What's this PR do?
Report back submissions will be retried up to 3 times if the response comes back with an error response of some sort. An error is either a response with status code 400, 408, 500 (these are just ones we've commonly seen elsewhere) or a body with a value == FALSE.

Other things:
- Logging the error when a response has a body == FALSE
- Updated the report back test to hopefully fix the random timeout issues we were seeing on tests

###### ds-content-api/index.js
Report back submissions are now done with the node-request-retry module. This allows us to retry failed requests and also set a custom callback (`isCampaignsReportBackError`) that determines if a response is valid or not.

###### reportback.test.js
The changes here make `createTestDoc` wait until the doc is created before continuing to the next step.

- `$ npm update`
- `$ npm test`
- Ensure you have appropriate values for `DS_CONTENT_API_*` in your environment to login and submit to the staging server.
- `$ node server.js`
- Use Postman to submit requests to test the report back flow.

#### Any background context?
It was seen a couple times that report backs submitted through the DS content API would return FALSE. But then subsequent submissions seemed to fix it. So this is sort of a brute-force way of making fixing it.

Close #402

### DIFF
--- a/app/lib/ds-content-api/index.js
+++ b/app/lib/ds-content-api/index.js
@@ -316,10 +316,10 @@ function campaignsReportback(rbData, callback) {
  */
 function isCampaignsReportbackError(response, body) {
   if (body && body.length > 0 && body[0] != false) {
-    return true;
+    return false;
   }
   else {
-    return false;
+    return true;
   }
 }
 
--- a/app/lib/ds-content-api/index.js
+++ b/app/lib/ds-content-api/index.js
@@ -1,6 +1,7 @@
 var request = require('request')
   , crypto = require('crypto')
   , logger = rootRequire('app/lib/logger')
+  , RequestRetry = require('node-request-retry')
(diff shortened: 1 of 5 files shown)


tongxiang:
No timeouts for me from reportback tests, incidentally.

tongxiang:
:+1:

PR #405, opened 2015-02-09
remove `endgame_config` from top-level config object, move into `story[END-GAME]`

tongxiang:
#### What's this PR do?
Today, I removed the `endgame_config` object from all SMS game documents on production, and moved the endgame config flags into the `END-GAME` object within the `story` object. 

To support this move, I made attendant changes to `gameMessageHelpers.js` (there are a lot of changes, mostly because I also changed the `storyConfig` variable to `gameConfig`; I reasoned that the variable should be called `gameConfig` so that the developer will not confuse that variable for the `story` object within a game's configuration object.) 

I also removed the `endgame_config` object from the corresponding model. 

After updating your local configuration database, (see the [wiki](https://github.com/DoSomething/ds-mdata-responder/wiki/Configuration) for a guide), running `npm test` should adequately prove that endgame logic is functioning normally. 

We're making this change to simplify development of the intertwine exporter/importer functions.

### DIFF
--- a/app/lib/sms-games/config/competitiveStoriesConfigModel.js
+++ b/app/lib/sms-games/config/competitiveStoriesConfigModel.js
@@ -53,17 +53,6 @@ var competitiveStoriesConfigSchema = new mongoose.Schema({
 
   },
 
-  // Flags configuring the game's endgame logic. 
-  endgame_config: {
-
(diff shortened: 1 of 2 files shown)


jonuy:
:+1:

PR #408, opened 2015-03-10
Config name refactoring

### LINKED ISSUE #389: create an enum of mongoose collection name strings to avoid hardcoding

jonuy:
#### What's this PR do?
This is a minor refactor that defines config model names in a single location at `app.ConfigName`. When getting configs now, instead of using hardcoded strings, we use object properties. Also changed some model names from their plural to singular forms.

- `smsConfigsLoader.js`: Model names defined here
- `*ConfigModel.js`: Plural names changed to singular
- Everything else: Changing from using hardcoded string to app.ConfigName

#### How did I test?
I tested by running `$ npm test` and sending requests to sms-multiplayer-game, reportback, and ds-routing endpoints to ensure they responded with valid responses.

Closes #389

@tongxiang

### DIFF
--- a/app/lib/ds-routing/config/startCampaignTransitionsConfigModel.js
+++ b/app/lib/ds-routing/config/startCampaignTransitionsConfigModel.js
@@ -23,5 +23,5 @@ var startCampaignTransitionsConfigSchema = new mongoose.Schema({
 })
 
 module.exports = function(connection) {
-  return connection.model('start_campaign_transition', startCampaignTransitionsConfigSchema);
+  return connection.model(app.ConfigName.CAMPAIGN_TRANSITIONS, startCampaignTransitionsConfigSchema);
 }
--- a/app/lib/ds-routing/config/startCampaignTransitionsConfigModel.js
+++ b/app/lib/ds-routing/config/startCampaignTransitionsConfigModel.js
@@ -23,5 +23,5 @@ var startCampaignTransitionsConfigSchema = new mongoose.Schema({
 })
 
 module.exports = function(connection) {
-  return connection.model('start_campaign_transitions', startCampaignTransitionsConfigSchema);
+  return connection.model('start_campaign_transition', startCampaignTransitionsConfigSchema);
 }
--- a/app/lib/ds-routing/config/yesNoPathsConfigModel.js
+++ b/app/lib/ds-routing/config/yesNoPathsConfigModel.js
@@ -25,5 +25,5 @@ var yesNoPathsConfigSchema = new mongoose.Schema({
 })
 
 module.exports = function(connection) {
-  return connection.model('yes_no_paths', yesNoPathsConfigSchema);
+  return connection.model('yes_no_path', yesNoPathsConfigSchema);
 }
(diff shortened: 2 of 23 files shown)


app/lib/ds-routing/config/startCampaignTransitionsConfigModel.js
```diff
@@ -23,5 +23,5 @@ var startCampaignTransitionsConfigSchema = new mongoose.Schema({
 })
 
 module.exports = function(connection) {
-  return connection.model('start_campaign_transitions', startCampaignTransitionsConfigSchema);
+  return connection.model('start_campaign_transition', startCampaignTransitionsConfigSchema);
```
tongxiang: was there a reason why you didn't change the string to `app.ConfigName.CAMPAIGN_TRANSITIONS` here?

app/lib/ds-routing/config/yesNoPathsConfigModel.js
```diff
@@ -25,5 +25,5 @@ var yesNoPathsConfigSchema = new mongoose.Schema({
 })
 
 module.exports = function(connection) {
-  return connection.model('yes_no_paths', yesNoPathsConfigSchema);
+  return connection.model('yes_no_path', yesNoPathsConfigSchema);
```
tongxiang: was there a reason why you didn't change the string to `app.ConfigName.YES_NO_PATHS` here?

app/lib/ds-routing/config/startCampaignTransitionsConfigModel.js
tongxiang: Ahh, yes. Because these are loaded before the app object is created.

tongxiang: Not true. This file, however, is loaded before the variable object is defined.

jonuy:
@tongxiang To your comments on why I still have the string in the `connection.model` calls, it was primarily because of this case here:

```
return connection.model('competitive_story', competitiveStoriesConfigSchema, 'competitive_stories');
```

I suppose I could've used `app.ConfigName.COMPETITIVE_STORIES` in place of the model name, but then I'd also need to provide the collection name. Do I have another property like `app.ConfigCollectionName`? Then I decided to just be lazy about it and leave it as is. So I'm not terribly opinionated about how it should go, and really not sure what would seem to be best. What do you think?

tongxiang:
Oh yeah I just puzzled over exactly this just now. 

It seems crappy to have different model and collection names. But I do think the benefit of being consistent, and using `app.ConfigName.xxxxxx` across the board for model names, outweighs the confusion with one model.

jonuy:
That's a fair point, I'll make that change.

tongxiang:
I can commit and push?

jonuy:
@tongxiang  Oh actually, yea, sure go for it.

PR #409, opened 2015-03-12
Reportback Caption

### LINKED ISSUE #406: Add caption question to report back flow
This could be done in one of two ways.

1. Replace/combine the "why" question with the caption question. This will only require us to modify the data that's sent to the Content API.
2. Add an extra question to ask for a caption for the photo. This will need both a modification to data sent to the Content API and support for an additional step in the report back flow.

Waiting on what data suggests we do. Should hear back by EOD 2/25.

cc: @mikefantini

jonuy:
#### Important!
We'll need to update the campaigns on Mobile Commons that are actively receiving report backs before pushing this to production. We should also test on staging before pushing to prod. Trello card to remind us is here: https://trello.com/c/CC8rbPaz/168-sms-report-backs-with-captions

#### What's this PR do?
During the report back flow, users will be asked an additional question to provide a photo caption. The order now goes: `photo -> caption -> quantity -> why -> complete`.

- `app/lib/ds-content-api/index.js`: adds a caption field to be submitted with the reportback request
- `app/lib/reportback/reportbackConfigModel.js`: adds a property for us to configure for the caption opt-in path
- `app/lib/reportback/index.js`: The meat of everything. Works pretty much like how we receive the quantity and send the next message to ask for why. Here we're just dropping in the caption message and receiving in between the photo receiving and quantity message.
- `app/lib/reportback/test/reportback.test.js`: tests modeled after the other tests already done

`$ npm test` and also tested by submitting to my local box with Postman.

Closes #406

### DIFF
--- a/app/lib/ds-content-api/index.js
+++ b/app/lib/ds-content-api/index.js
@@ -283,6 +283,7 @@ function campaignsReportback(rbData, callback) {
   var nid = rbData.nid;
   var body = {
     uid: rbData.uid,
+    caption: rbData.caption,
     quantity: rbData.quantity,
     why_participated: rbData.why_participated,
     file_url: rbData.file_url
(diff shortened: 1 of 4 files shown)


tongxiang:
:+1:

PR #421, opened 2015-04-03
Update Mobile Commons API integration to facilitate staff pick functionality across affiliate accounts

### LINKED ISSUE #413: Support mobilecommons.optout with a different affiliate account key
We're starting to setup mDatas in more Mobile Commons affiliate accounts than just the main one. So things like campaign transitions need to work which involves doing a `mobilecommons.optout` from a campaign.

The problem is that call needs the company key which is a value that differs from one affiliate account to another.

Maybe add an optional parameter to include the company key in the `optout` request? Or maybe a parameter that indicates what affiliate account to use, and have that data structure stored within the `mobilecommons` module itself?

tongxiang:
#### What's this PR do?
Updates our usage of the Mobile Commons "optout" API so that the API call no longer requires a "company key" param. This facilitates the normal staff pick flow when a user is opted out of the campaign signup flow and opted into the start campaign flow. 

Also updates reportback and mobilecommons tests. 

Tested with the main DoSomething.org affiliate account by testing with personal phone. Texted "WORN" to opt into the Comeback Clothes 2015 signup campaign. Then, opted out of that signup campaign and into the main start campaign flow with [the DS Start Campaign Post in this Postman Collection.](https://www.getpostman.com/collections/468ccfa1dd5e91d672a3) Then checked my phone's profile on MobileCommons to ensure I've successfully been opted out of the signup campaign. 

Still needs to be tested with a campaign on the affiliate account, but none have been set up yet. Will test again when @ArielSD sets it up next week. 

Closes #413.

### DIFF
--- a/app/lib/ds-routing/controllers/MCRouting.js
+++ b/app/lib/ds-routing/controllers/MCRouting.js
@@ -53,7 +53,7 @@ MCRouting.prototype.yesNoGateway = function(request, response) {
 };
 
 /**
- * Transition users for the sign up campaign to the actual campaign.
+ * Transition users from the sign up campaign to the actual campaign.
  */
 MCRouting.prototype.campaignTransition = function(request, response) {
   if (typeof(request.body.mdata_id) === 'undefined') {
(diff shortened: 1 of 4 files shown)


jonuy:
:+1:

PR #422, opened 2015-04-03
fix beta opt in profile update

### LINKED ISSUE #420: Re-implement playerNameHelpers.createGameInviteAll() method
To resurrect the player name code I commented out in #416.

Code prior to #416 was breaking beta tell-a-friend functionality. But then #416 took out the code that allowed some messages with player names to work.

tongxiang:
#### What's this PR do?
Code prior to #416 was breaking beta tell-a-friend functionality. But then #416 took out the code that allowed some messages with player names to work. We're reconciling everything. 

Now, messages with player names work, and beta tell-a-friend functionality still works. 

We've added a callback to the `mobilecommons.profile_update` function to make everything work. 

Use [this Postman collection](https://www.getpostman.com/collections/4331949cf475d70cf3db
) to test, perhaps with phone numbers you have access to. Create a game, ensure that the alpha gets a message with the names of the users she's invited, and when betas join, ensure that the proper betas get the right messages. 

I've modified the `Science Sleuth A/B testing` game so that it supports this new feature. This postman collection creates a new game within that campaign. 

Closes #420 

#### Questions:
Is adding an optional callback inelegant? Is there a better way of doing it?

### DIFF
--- a/app/lib/sms-games/controllers/playerNameHelpers.js
+++ b/app/lib/sms-games/controllers/playerNameHelpers.js
@@ -20,11 +20,10 @@ module.exports = {
  * For the purposes of SMS games, that message will contain those names. 
  * If names aren't present, phone numbers are used. 
  *
+ * @param gameConfig
+ *  The configuration document for the game, containing opt in path ids. 
  * @param  gameDoc
  *  The document storing game data, including player names and numbers. 
- * @param alphaWaitOip
- *  The opt in path of the message sent to the alpha alerting her that she's
- *  created a game. Contains the names of invited players. 
  * 
  */
 function createGameInviteAll(gameConfig, gameDoc) {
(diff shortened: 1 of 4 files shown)


jonuy:
The callback here feels weird because it's not available with any of the other `mobilecommons` calls and in this case it only executes when a response comes back successfully.

What might be easier is adding custom fields to the `mobilecommons.optin` call. The [docs](https://mobilecommons.zendesk.com/hc/en-us/articles/202052484-Web-forms) suggest that we can push custom fields along with the `join` call by putting it in something like `person[players_not_in_game]`.

tongxiang:
Great point @jonuy. Modified the `mobilecommons.optin` call to support the addition of custom fields for the alphPhone user. (Also did some general refactoring of the `exports.optin` function, and added underscore to the `mobilecommons.js` test file to allow for the determination of an object's deep equality.)

jonuy:
:+1:

Issue #443, opened 2015-05-05
add test coverage for #442


tongxiang:
No interference between a new reportback and an unfinished one.

tongxiang:
Closed by #444.

PR #444, opened 2015-05-05
fix bug - unfinished reportback influencing new reportback flow, add tests

### LINKED ISSUE #442: Reportbacks asking for a caption will incorrectly reply with the error message that it's not a photo
Steps:
1. Start a reportback flow on SMS
2. Submit photo when asked for one
3. Submit caption when asked for one
4. See that the system replies back with the error message that it's not a photo

I haven't figured out yet how to consistently reproduce this bug.

The working assumption is that this comes from starting one reportback (A) and not completing it. Then starting another reportback (B) and trying to complete that reportback B. I'm not sure right now how far down A you have to go in order for the bug to show up on B.

tongxiang:
#### What's this PR do?
Previously, if a reportback DB document was created but the reportback was never submitted/finished by the user, it was never destroyed and thus lingered in the database. 

If a user didn't finish a reportback flow, but then wanted to reportback into another campaign, our app would sometimes find their previous unfinished reportback documents and update those documents, messing up the new reportback flow. 

This PR adds a new, campaign-specific param to our DB calls to eliminate this interference. 

`reportback/index.js` contains the modified code. 

By beginning a reportback flow (1/3 Postman collection [here](https://www.getpostman.com/collections/71bd60b99e46cc8ce514
)), stopping at any stage before you submit the reportback, and then creating a new reportback flow. (Comeback Clothes Postman collection [here](https://www.getpostman.com/collections/8ce30f8111deea794ba9)). 

Created #443 to add test coverage. 

Closes #442, #443.

### DIFF
--- a/app/lib/reportback/test/reportback.test.js
+++ b/app/lib/reportback/test/reportback.test.js
@@ -8,24 +8,23 @@ var assert = require('assert')
 function test() {
   var TEST_PHONE = '15555555555'
     , TEST_CAMPAIGN_CONFIG = app.getConfig(app.ConfigName.REPORTBACK, 'test', 'endpoint')
-    , testDoc // exists in the global test scope; variable re-assigned each time createTestDoc() is run. 
-    ; 
+    , TEST_CAMPAIGN_CONFIG_2 = app.getConfig(app.ConfigName.REPORTBACK, 'test2', 'endpoint')
+    ;
 
-  var createTestDoc = function(done) {
-    model.create({phone: TEST_PHONE, campaign: TEST_CAMPAIGN_CONFIG.endpoint}, function(err, doc) {
-      if (doc) {
-        testDoc = doc;
-        done();
-      }
-    });
-  };
-
-  var removeTestDoc = function() {
+  var removeTestDocs = function() {
     model.remove({phone: TEST_PHONE}).exec();
   };
 
   describe('reportback.findDocument', function() {
-    before(createTestDoc);
+    var testDoc;
+    before(function(done) {
+      model.create({phone: TEST_PHONE, campaign: TEST_CAMPAIGN_CONFIG.endpoint}, function(err, doc) {
+        if (doc) {
+          testDoc = doc;
+          done();
+        }
+      });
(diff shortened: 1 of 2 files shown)


app/lib/reportback/test/reportback.test.js
```diff
@@ -6,12 +6,15 @@ var assert = require('assert')
   ;
 
 function test() {
-  var TEST_PHONE = '15555555555';
-  var TEST_CAMPAIGN_CONFIG = app.getConfig(app.ConfigName.REPORTBACK, 'test', 'endpoint');
+  var TEST_PHONE = '15555555555'
+    , TEST_CAMPAIGN_CONFIG = app.getConfig(app.ConfigName.REPORTBACK, 'test', 'endpoint')
+    , testDoc // exists in the global test scope; variable re-assigned each time createTestDoc() is run.
```
jonuy: Instead of pulling this var up to the global scope, I think I'd prefer we continue using the `testDoc` vars in their previously more local scope and just add the campaign property. **Or** to pass in `testData` to both parameters - ex: `reportback.receivePhoto(testData, testData);`

Having this particular mutable variable in the global scope can be confusing without context of the entire setup of tests in this file. Like, I can see `before(createTestDoc);` in a test, but not necessarily know intuitively that the `testDoc` var has now just been changed.

tongxiang:
Changed the global `testDoc` document to be local to each test, added tests for the situation where a user submits one reportback, doesn't finish it, and then begins another. @jonuy

app/lib/reportback/test/reportback.test.js
```diff
@@ -337,12 +371,143 @@ function test() {
...
+          if (doc && doc.quantity == quantityData.args) {
+            testDoc2 = doc;
+            done();
+          }
+        })
+      })
+      reportback.handleUserResponse(testDoc2, quantityData);
+    })
+
+    after(function() {
```
jonuy: Seems like this could just be `after(removeTestDocs);`

jonuy:
:+1:

PR #468, opened 2015-06-09
remove new relic 

tongxiang:
#### What's this PR do?
Removes the traditional new relic call from this file, since it's baked into Heroku's deployment.


aaronschachter:
This was the last instance of using the `NEWRELIC_LICENSE_KEY` -- we instead use a `NEW_RELIC_LICENSE_KEY` (#536). Removing the `NEWRELIC_LICENSE_KEY` from our config variables to help Heroku support troubleshoot lack of errors in https://github.com/DoSomething/ds-mdata-responder/issues/518#issuecomment-238294705

PR #471, opened 2015-07-13
Issue 470: reportback quantity validations

### LINKED ISSUE #470: add validation of numbers to reportback submission
We don't want users to be able to submit 0-valued reportbacks. We also want to be able to convert numbers spelled out to integers (i.e., four hundred --> 400).

tongxiang:
#### What's this PR do?
Adds a [library](https://github.com/DoSomething/count-von-count) which filters a text string for numbers and spelled numbers, and converts those numbers to digits. 

This library is added to the reportback flow, where the user-submitted response for the "quantity done" question is then filtered for digits before being submitted to the reportback API. 

If a response with no numbers is submitted, the user receives an error message, which has been added to the reportback config do. 

`npm test` validates whether the user receives the `message_quantity_sent_invalid` message after submitting an invalid message. 

The [library](https://github.com/DoSomething/count-von-count) contains its own tests, and its validity can be tested independently. 

This functionality can also be tested with [this postman collection](https://www.getpostman.com/collections/2abbad22b2d79005366e). 

`npm test`. 

When receiving a reportback submission through its API, the Drupal DoSomething app doesn't accept anything other than digits--"1", "2"--for the quantity of things done. 

Previously, if a user submitted a non-digit response during the "quantity" question of the SMS reportback flow--i.e., if a user submitted "one" instead of "1"--the DoSomething app would interpret that as "0". 

This became problematic during campaigns like "I Heart Dad". When users were asked "how many dads did you measure the blood pressure", a large percentage of users responded with "the only one I have" or "one". 

Closes #470.

### DIFF
--- a/app/lib/reportback/index.js
+++ b/app/lib/reportback/index.js
@@ -13,6 +13,7 @@ var express = require('express')
   , REPORTBACK_PERMALINK_BASE_URL
   , shortenLink = rootRequire('app/lib/bitly')
   , Q = require('q')
+  , parseForDigits = require('count-von-count')
   ;
 
 if (process.env.NODE_ENV == 'production') {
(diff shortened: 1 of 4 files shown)


jonuy:
:+1:

Issue #476, opened 2015-07-23
Limit users to only 5 donations


jonuy:
This is particularly for the upcoming Science Sleuth campaign, but should be usable for any campaign that we do donation flows with.

Potential content needs:
- [ ] message indicating that they've reached the maximum number of times they can donate

PR #482, opened 2015-07-28
Limits users to max 5 donations

jonuy:
#### What's this PR do?

Users will be able to make donations for a single campaign up to 5 times (or whatever is set in the config doc). The configuration for a donation flow now includes a `max_donations_allowed` and `max_donations_reached_oip` property. The first sets the number of donations a single user is allowed to make. The second provides the Mobile Commons opt-in path that a user should be sent to if they attempt to donate past the maximum amount.
- **DonorsChooseDonationController.prototype.start:** Checks if the user has reached the max donation limit.
- **DonorsChooseDonationController.prototype.submitDonation:** Refactored a bit to move the code that's immediately run in this function to the top. All the closure functions are then listed underneath.
- **updateUserWithDonation** and **incrementDonationCount** are the functions that handle updating a user's donation history.
- **sgUser.js:** user model stores the donation count
- **donorschooseConfigModel.js** config properties added for handling max donations
#### How did I test?

Tested through Postman simulating the requests Mobile Commons would make to our app through mDatas.
1. /donations/donors-choose/start?id=[test id]
2. /donations/donors-choose/retrieve-location?id=[test id]
3. /donations/donors-choose/retrieve-firstname?id=[test id]
4. /donations/donors-choose/retrieve-email?id=[test id]
#476

### DIFF
--- a/app/lib/donations/controllers/DonorsChooseDonationController.js
+++ b/app/lib/donations/controllers/DonorsChooseDonationController.js
@@ -27,7 +27,7 @@ var donorsChooseApiKey = (process.env.DONORSCHOOSE_API_KEY || null)
 if (process.env.NODE_ENV == 'test') {
   donorsChooseApiKey = 'DONORSCHOOSE';
   donorsChooseApiPassword = 'helpClassrooms!';
-  donorsChooseDonationBaseURL = 'https://apiqasecure.donorschoose.org/common/json_api.html?APIKey=';
+  donorsChooseDonationBaseURL = 'https://dev1-apisecure.donorschoose.org/common/json_api.html?APIKey=';
 }
 
 var TYPE_OF_LOCATION_WE_ARE_QUERYING_FOR = 'zip' // 'zip' or 'state'. Our retrieveLocation() function will adjust accordingly.
--- a/app/lib/donations/controllers/DonorsChooseDonationController.js
+++ b/app/lib/donations/controllers/DonorsChooseDonationController.js
@@ -108,11 +108,11 @@ DonorsChooseDonationController.prototype.start = function(request, response) {
 
     // If user has not hit the max limit, start the donation flow.
     if (!config.max_donations_allowed || donationsCount < config.max_donations_allowed) {
-      sendSMS(phone, config.max_donations_reached_oip);
+      sendSMS(phone, config.start_donation_flow);
     }
     // Otherwise, send an error message
     else {
-      sendSMS(phone, config.start_donation_flow);
+      sendSMS(phone, config.max_donations_reached_oip);
     }
   }
 };
--- a/app/lib/donations/controllers/DonorsChooseDonationController.js
+++ b/app/lib/donations/controllers/DonorsChooseDonationController.js
@@ -315,7 +315,18 @@ DonorsChooseDonationController.prototype.retrieveEmail = function(request, respo
  *
  */
 DonorsChooseDonationController.prototype.submitDonation = function(apiInfoObject, donorInfoObject, proposalId, donationConfig) {
-  // First request: obtains a unique token for the donation.
+  var donorPhone;
+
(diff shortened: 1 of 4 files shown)


app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -58,6 +65,59 @@ function DonorsChooseDonationController() {
...
+
+    config = app.getConfig(app.ConfigName.DONORSCHOOSE, configId);
+
+    // If user has not hit the max limit, start the donation flow.
+    if (!config.max_donations_allowed || donationsCount < config.max_donations_allowed) {
+      sendSMS(phone, config.max_donations_reached_oip);
+    }
+    // Otherwise, send an error message
+    else {
+      sendSMS(phone, config.start_donation_flow);
```
tongxiang: Hypothetical case: 

If a user hasn't played the game before and thus has no user doc, but wants to make a donation. 

If that user texts in the `DONATE` keyword, they' would get trapped in this flow--since no user doc is found, the `start_donation_flow` error message would be sent. 

How should we handle? Should we just create a user doc for them? (To allow people who haven't played the game to donate?)

app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -58,6 +65,59 @@ function DonorsChooseDonationController() {
...
+        if (configId == doc.donations[i].config_id) {
+          donationsCount = doc.donations[i].count;
+          break;
+        }
+      }
+    }
+
+    config = app.getConfig(app.ConfigName.DONORSCHOOSE, configId);
+
+    // If user has not hit the max limit, start the donation flow.
```
tongxiang: shouldn't this comment be reversed with the one below it?

app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -58,6 +65,59 @@ function DonorsChooseDonationController() {
...
+    if (!config.max_donations_allowed || donationsCount < config.max_donations_allowed) {
```
tongxiang: shouldn't this be `donationsCount > config.max_donations_allowed`?

app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -58,6 +65,59 @@ function DonorsChooseDonationController() {
...
+      sendSMS(phone, config.max_donations_reached_oip);
+    }
+    // Otherwise, send an error message
```
tongxiang: this comment should be reversed with the one above it.

jonuy: oops, good catch. The logic is actually backwards here. hm... I should test again.

app/lib/donations/controllers/DonorsChooseDonationController.js
jonuy: yea, the opt in path I should be sending here is `config.start_donation_flow`. The logic in this `if` should be: if a `max_donations_allowed` has not been configured or if this person hasn't donated to the max limit yet, then allow it.

app/lib/donations/controllers/DonorsChooseDonationController.js
jonuy: Fixed by the logic bug you pointed out below. I confirmed this is ok now by testing with a user that had no user doc in the database yet.

jonuy:
@tongxiang thanks for pointing out the problem. Pushed a fix. Anything else you saw that might be problematic?

jonuy:
@tongxiang Rebased with master so I got the fix you put in with d530a437347d108990e5c2dc8d9661a9d9304861. The error now at least is back to the good ole `current_game_id` error we're seeing in other tests. I'm gonna merge this in and work on test bug in a separate branch.

PR #485, opened 2015-08-12
Test fix

tongxiang:


tongxiang:
Was trying to test a change against wercker.

PR #490, opened 2015-08-27
Skipping end level success/failure messages

jonuy:
#### What's this PR do?

At the end of a level, if no success/failure messages are defined in a game's config, then skip it.

Also includes a bug fix where it seemed like end level messages would get sent to betas who didn't accept the invite to a game
#### How was this tested?

This was tested with two players in a ScienceSleuth2015 game. I tried both with the `end_level_*` properties configured and without it. With them defined, it sent the end level success/failure messages. Without them, it didn't and sent just the group message instead.
#### Any background context?

https://trello.com/c/YGeIyhC9/249-deactivate-the-messaging-associated-with-these-oips-for-end-level-feedback-see-description-for-links

ScienceSleuth2015 content has been setup so that the failure/success names are built into the group end level message
#### TODOs

We'll need to update the production db to unset `end_level_0_correct_loss`, `end_level_solo_correct_win`, `end_level_1_to_4_correct_win`, and `end_level_1_correct_loss` on the ScienceSleuth2015 config.

### DIFF
--- a/app/lib/sms-games/controllers/playerNameHelpers.js
+++ b/app/lib/sms-games/controllers/playerNameHelpers.js
@@ -126,12 +126,6 @@ function endLevelMessageWithSuccessPlayerNames(gameConfig, gameDoc, delay) {
     return;
   }
 
-  // If no end_level_* messages are specified, then return.
-  if (!gameConfig.end_level_0_correct_loss && !gameConfig.end_level_1_correct_loss &&
-      !gameConfig.end_level_solo_correct_win && !gameConfig.end_level_1_to_4_correct_win) {
-    return;
-  }
-
   var i, j
     , path
     , alphaPlayer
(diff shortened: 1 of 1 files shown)


jonuy:
@tongxiang Hm, so I just tried something else out on my local. It also seems to properly skip the failure/success messages when we just remove the `end_level_*` properties from the config doc and not use the code changes I've put in this PR.

Think there's any value in merging this in anymore? Or maybe just the bug fix and not the `if` check at the top of the function?

tongxiang:
@jonuy since we've decided we're keeping in the end level player-name messages for at least the first week of the launch, we'll hold on this.

jonuy:
@tongxiang So this most recent change (074defb) adds a small bit of logic so that instead of trying to do an optin, we do a `mobilecommons.profile_update` when no `endLevelMessage` has been set in the config. It'll then allow for profile fields to still get updated and ultimately resolve this: https://trello.com/c/YGeIyhC9

Let me know if I'm not thinking through this fully, but I think this is safe to merge in since the configs on production right now all have those end level success/fail paths set. So this new bit of logic should never get hit. And even if it does... it doesn't seem harmful?

If merged, I plan on testing with a duplicate Science Sleuth config sans the end level paths.

tongxiang:
:+1:

PR #516, opened 2016-01-07
Adds a save_to and config_override query params to the reportback flow

jonuy:
#### What's this PR do?

When requests are made to `POST /reportback/:campaign_name`, we can now add a `save_to` query parameter. This allows us to specify what reportback field the user response should be saved into - even if it's already got a value in the reportback doc.

ex: `POST /reportback/:campaign_name?save_to=quantity`

The valid `save_to` values are:
- photo
- caption
- quantity
- why_important

**Update!** We also need to add a `config_override` query param to the request calls for this second custom reportback. This was needed because the user reportback docs that exist are already keyed up with the original campaign `endpoint` value. In order to distinguish between the original reportback config and the new one we'll create, we'll specify it by this `config_override`. Am I making sense? It's the end of the day and I just wanna get this thing done and out the door.

Requests come in through the `router.post`. The `save_to` value is added to the custom object that gets passed to `handleUserResponse()`. And then the logic in there determines how it should process the response and save it.
#### What was tested?

I manually tested two scenarios on my local setup:
1. User filled out multiple fields for the reportback - photo, caption, quantity. Then using `?save_to=`, I verified that I could overwrite the values for each
2. User submitted only a photo. Typically, the next response should be put into caption, but using `?save_to=quantity` was able to send the next value into the `quantity` field.
#### Any background context?

This is primarily needed right now for a custom reportback for users who have submitted a photo but didn't get to complete the rest of the reportback afterwards.

### DIFF
--- a/app/lib/reportback/index.js
+++ b/app/lib/reportback/index.js
@@ -23,6 +23,24 @@ else {
   REPORTBACK_PERMALINK_BASE_URL = 'http://staging.beta.dosomething.org/reportback/';
 }
 
+/**
+ * Single endpoint through which a reportback is created, populated and
+ * submitted. The expected reportback flow is that a user is asked for a photo,
+ * a caption, the reportback quantity and finally why it's important to them.
+ *
+ * POST /reportback/:campaign
+ *
+ * Query Params:
+ *   save_to - (optional) force this response to save to the specified field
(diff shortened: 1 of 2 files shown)


jonuy:
Aiming to get a broadcast out today that will rely on this code. Gonna push and deploy and test on Mobile Commons.

Issue #518, opened 2016-01-15
Fix NewRelic monitoring


jonuy:
*Updated 8/8*
* [x] Get error logging working

*Updated 7/19 -- next steps per https://github.com/DoSomething/ds-mdata-responder/issues/518#issuecomment-233766089*
* [x] Remove the New Relic add-on
* [x] Remove the custom NR code (require('newrelic');)
* [x] Verify we have no more NR-related log activity
* [x] Re-introduce the newer NR package in code

*Original description*

I'm not sure when it stopped, but it seems like we're no longer getting any reporting to our NewRelic account from this app.

Initially thought that it may be because `require('newrelic')` isn't the first line in [server.js](https://github.com/DoSomething/ds-mdata-responder/blob/dev/server.js#L8) like the NewRelic docs ask for it to be. But that change was done in late 2014 and I'm _pretty_ sure we didn't just go through the entirety of 2015 without getting any NewRelic reporting.

aaronschachter:
The [logs on production](https://dashboard.heroku.com/apps/ds-mdata-responder/logs) are flooded with New Relic related messages, making it really hard to spot any other errors.

> 2016-07-18T15:09:19.214273+00:00 app[web.1]: {"v":0,"level":10,"name":"newrelic","hostname":"9e24e984-e7c8-4f7b-8cc5-f72b27e63081","pid":14,"time":"2016-07-18T15:09:19.213Z","msg":"Not creating segment \"timers.setTimeout\" because no transaction was active","component":"tracer"}
2016-07-18T15:09:29.528610+00:00 app[web.2]: {"v":0,"level":10,"name":"newrelic","hostname":"eb8c319a-467a-4c85-b398-8c01785579b4","pid":14,"time":"2016-07-18T15:09:29.528Z","msg":"Not tracing MongoDB unknown._executeQueryCommand(); no New Relic transaction.","component":"mongodb"}
2016-07-18T15:09:29.528973+00:00 app[web.2]: {"v":0,"level":10,"name":"newrelic","hostname":"eb8c319a-467a-4c85-b398-8c01785579b4","pid":14,"time":"2016-07-18T15:09:29.528Z","msg":"Not tracing MongoDB unknown._executeQueryCommand(); no New Relic transaction.","component":"mongodb"}

cc @blisteringherb

aaronschachter:
[...]

PR #530, opened 2016-07-19
Removes NewRelic package

aaronschachter:
#### What's this PR do?
Removes the `newrelic` package to cut down on some [New Relic messages](https://github.com/DoSomething/ds-mdata-responder/issues/518#issuecomment-233358812) we're seeing in production:
>  {"v":0,"level":10,"name":"newrelic","hostname":"Aarons-MacBook-Pro-2.local","pid":1029,"time":"2016-07-19T21:13:04.305Z","msg":"Not creating segment \"timers.setTimeout\" because no transaction was active","component":"tracer"}

Run `heroku local` before and after to verify this branch stops the `newrelic` logs.

Deploying to prod will complete step 2 in #518

### DIFF
--- a/server.js
+++ b/server.js
@@ -3,11 +3,6 @@ global.rootRequire = function(name) {
   return require(__dirname + '/' + name);
 }
 
-// Nodejitsu specifies NODE_ENV='production' by default
-if (process.env.NODE_ENV == 'production') {
-  require('newrelic');
-}
-
 var express = require('express')
   , path = require('path')
   , http = require('http')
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -3,11 +3,6 @@ global.rootRequire = function(name) {
   return require(__dirname + '/' + name);
 }
 
-// Nodejitsu specifies NODE_ENV='production' by default
```
aaronschachter: This may help explain the 2 New Relic integrations we're seeing in https://github.com/DoSomething/ds-mdata-responder/issues/518#issuecomment-233755953.

PR #535, opened 2016-07-21
Include reportback.source when posting to Phoenix API

### LINKED ISSUE #519: Add a `source` field to reportback submissions
Similar to how we'd add a `source` field to Phoenix registration calls, we should do the same for reportbacks. This is to make it easier for us on the data side to determine what platform reportbacks are coming from.

aaronschachter:
Fixes #519


aaronschachter:
Discussed with Quasar team, we're going to post `sms-mobilecommons` as our Reportback source.

PR #568, opened 2016-08-09
Science Sleuth Donations: collect zip and filter Projects query

### LINKED ISSUE #566: Collect zip code in Science Sleuth donation flow
Refs #539: Change the `DonorsChooseDonationController` flow to first collect the member's zip code, and store in Mobile Commons appropriately. Prompt for first name next.

The old implementation of this feature would additionally validate if the input is a valid zip code, and opt the member into an `invalid_zip_oip` if not. see #191 

We're collecting zip instead of geo-targetting (#480) to cut down on moving parts (needing to query out to an external service for zip by coordinates)

## Tasks
* [ ] Modify config (or set environment variable) to store Invalid Zip OIP
* [ ] Restore collecting/store user's zip and opt into Retrieve First Name OIP / Invalid Zip OIP
* [ ] Update documentation

aaronschachter:
#### What's this PR do?
* Closes #566: Renames `retrieveLocation` function as `retrieveZip`, and changes it to call `findProject` with new param for `zip` to filter projects query by (instead of posting to itself with the `_post` function).
* Removes deprecated `/donations/donors-choose/find-projects` route and related `DonorsChooseDonationController.post()` function and `DonorsChooseDonationController.host` property
* Removes deprecated retrieve projects by state functionality. Less code, less moving parts
* Changes the `start` function to opt user into the `oip_ask_zip` if `max_donations_allowed` has not been reached. Previously the `start` function would simply call `findProject` without filtering for zip.
* For now, uses a hardcoded `donorsChooseConfig` variable to DRY instead of constantly reading calling `app.getConfig` in every function. Also stores donation count against the MoCo Campaign ID instead of the `config_id`. See todos section
    * Renames properties with `oip_` prefix
* Replaces the `sendSMS` helper function with the `mobilecommons.profile_update` function it calls, to better indicate how all this stuff works

Pull down and POST to the following endpoints on your localhost in this order (using your mobile number as a `phone` body parameter to verify correct messages sent back from MoCo), and confirming that zip code is collected and the selected project is filtered accordingly:
* `/donations/donors-choose/start`
* `/donations/donors-choose/retrieve-zip`
* `/donations/donors-choose/retrieve-firstname`
* `/donations/donors-choose/retrieve-email`

Hardcoded the `donorsChooseConfig` oip's to avoid making changes to `donorschooseConfigModel.js` for right now. 2014 and 2015 both had different schema, 2016 now needs an Invalid Zip OIP which was missing from the file (removed in 2015):

2014 vs 2015

I'm not really sold on staying with `config` documents for these, as the ID's 101 and 201 are just arbitrarily chosen as far as I can tell. It seems logical to use the Moco Campaign ID instead. I've been toying with the idea of storing the actual message copy in the `donorsChooseConfig ` object instead of opt-in-paths, and saving the relevant response to a custom field like we do in the `/slothbot/mobilecommons` endpoint (#537). e.g `msg_invalid_zip`, `msg_ask_email` and storing to a `ss_response` Custom Field in Mobile Commons.

#### todo
* [ ] Resolve hardcoded opt-in paths, either sticking with the `config` documents and updating our `donorschooseConfigModel` accordingly, or setting as environment variables instead
    * [ ] Create clone of MoCo Donation 2016 campaign and mData's to have staging/production
    * [ ] Store user donations by `config_id` or MoCo Campaign ID based on decision
* [ ] Add checks for `request.data.phone` and `request.data.args` when applicable in each endpoint
* [ ] Update API documentation for `retrieve-zip` endpoint
* [ ] Update [ServerConfig:DonorsChoose](https://github.com/DoSomething/ServerConfig/wiki/4.2-DonorsChoose) to document the endpoints 2015 mData's pointed to (i've overwritten them because we will never use them again, and have 191 mData's in MoCo - majority exist never to be used again)

### DIFF
--- a/app/lib/donations/controllers/DonorsChooseDonationController.js
+++ b/app/lib/donations/controllers/DonorsChooseDonationController.js
@@ -48,9 +48,19 @@ var connectionOperations = rootRequire('app/config/connectionOperations')
   , userModel = rootRequire('app/lib/sms-games/models/sgUser')(connectionOperations)
   ;
 
-// @TODO: add some kind of reference to retrieve the donations config 
-// var donationConfigModel = require('../')
-var donationConfigId = (process.env.DONORSCHOOSE_CONFIG_ID || 201);
+var donorsChooseConfig = {
+  max_donations_allowed: 5,
+  moco_campaign: 146307,
+  oip_ask_email: 209781,
+  oip_ask_first_name: 209779,
+  oip_ask_zip: 211113,
+  oip_donation_complete_1: 209783,
+  oip_donation_complete_2: 209785,
+  oip_donation_complete_3: 209787,
+  oip_error: 209791,
+  oip_invalid_zip: 211115,
+  oip_max_donations: 209789
+};
 
 function DonorsChooseDonationController() {
   this.host; // Used to store reference to application host.
@@ -92,23 +102,19 @@ DonorsChooseDonationController.prototype.start = function(request, response) {
     donationsCount = 0;
     if (doc && doc.donations) {
       for (i = 0; i < doc.donations.length; i++) {
-        // Check for number of donations from this specific Donation Flow config/campaign.
-        if (donationConfigId == doc.donations[i].config_id) {
+        // Check for number of donations from this specific Donation Flow MoCo Campaign.
+        if (donorsChooseConfig.moco_campaign == doc.donations[i].config_id) {
           donationsCount = doc.donations[i].count;
           break;
         }
       }
     }
 
-    config = app.getConfig(app.ConfigName.DONORSCHOOSE, donationConfigId);
-
-    // If user has not hit the max limit, start the donation flow.
-    if (!config.max_donations_allowed || donationsCount <= config.max_donations_allowed) {
-      self.findProject(phone, config);
+    if (donationsCount <= donorsChooseConfig.max_donations_allowed) {
+      self.findProject(phone);
     }
-    // Otherwise, send an error message
     else {
-      sendSMS(phone, config.max_donations_reached_oip);
+      sendSMS(phone, donorsChooseConfig.oip_max_donations);
     }
   }
 };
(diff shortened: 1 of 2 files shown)


app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -165,7 +168,7 @@ DonorsChooseDonationController.prototype.findProject = function(mobileNumber, co
         };
       }
       else {
-        sendSMS(mobileNumber, config.error_direct_user_to_restart);
```
aaronschachter: This `error_direct_user_to_restart ` property was removed in the 2015 schema and would have caused an error.

app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -187,11 +190,11 @@ DonorsChooseDonationController.prototype.findProject = function(mobileNumber, co
       // their email address and attempts to find the document to be updated. 
       donationModel.create(currentDonationInfo).then(function(doc) {
         logger.log('debug', 'DonorsChoose.findProject created donation_infos document:%s for user:%s', JSON.stringify(doc), mobileNumber);
-        mobilecommons.profile_update(mobileNumber, config.start_donation_flow, mobileCommonsCustomFields); // Arguments: phone, optInPathId, customFields.
+        mobilecommons.profile_update(mobileNumber, donorsChooseConfig.oip_ask_first_name, mobileCommonsCustomFields);
       }, promiseErrorCallback('DonorsChoose.findProject unable to create donation_infos document for user: ' + mobileNumber));
     }
     else {
-      sendSMS(mobileNumber, config.error_direct_user_to_restart);
```
aaronschachter: Same here - `error_direct_user_to_restart ` property was removed in the 2015 schema and would have caused an error.

app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -483,124 +480,54 @@ DonorsChooseDonationController.prototype.submitDonation = function(donorInfoObje
...
-  });
-}
-
-/**
- * Sets the hostname.
- *
- * @param host
- *   The hostname string.
- */
-DonorsChooseDonationController.prototype.setHost = function(host) {
```
aaronschachter: RIP double declaration of `DonorsChooseDonationController.prototype.setHost`  💐

app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -483,124 +480,54 @@ DonorsChooseDonationController.prototype.submitDonation = function(donorInfoObje
...
  * @param request
  *   Express Request object
  * @param response
  *   Express Response object
  */
-DonorsChooseDonationController.prototype.retrieveLocation = function(request, response) {
-
+DonorsChooseDonationController.prototype.retrieveZip = function(request, response) {
   if (typeof request.body.phone === 'undefined' || typeof request.body.args === 'undefined') {
     response.status(406).send('Missing required params.');
```
aaronschachter: Refs #569

Issue #552, opened 2016-08-03
Science Sleuth extras


aaronschachter:
Running list of nice to have's if we get time.

* [x] If user is donating for 2nd time, we shouldn't ask for any of their donation info again because we already have it saved

* [x] Set a Donation Config ID environment variable to use instead of hardcoding ID into each mData, which adds more opportunities for config errors

* [x] Store a running count of completed donations to keep an eye on how much $ is left in our DonorsChoose account.
    * Not handled in code, campaigns lead will receive daily balance emails from Donors Choose

aaronschachter:
`DONORSCHOOSE_CONFIG_ID` was added in #556

PR #571, opened 2016-08-10
DonorsChooseDonationController.getDonationCount

aaronschachter:
#### What's this PR do?
Extracts the code that inspects a `sg_users` document for the user's total of successful donations for the current Donation campaign into its own function, `getDonationCount(userDocument)`. 

Pull down and test the `/donations/donors-choose/start` endpoint locally to verify nothing's changed.

Ideally the `start` endpoint shouldn't ask the user for their information again if this is their 2nd donation (#552).  Moving the donation count check into its own function helps make the `start` function more readable, especially if we'll be adding more logic into it.

#### Relevant tickets
None

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/lib/donations/controllers/DonorsChooseDonationController.js
+++ b/app/lib/donations/controllers/DonorsChooseDonationController.js
@@ -73,6 +73,20 @@ function DonorsChooseDonationController() {
 DonorsChooseDonationController.prototype.resourceName = 'donors-choose';
 
 /**
+ * For given user document, return # of donations completed for our current Donation campaign.
+ */
+function getDonationCount(userDocument) {
+  if (userDocument && userDocument.donations) {
+    for (var i = 0; i < userDocument.donations.length; i++) {
+      if (donorsChooseConfig.moco_campaign == userDocument.donations[i].config_id) {
+        return userDocument.donations[i].count;
+      }
+    }
+  }
+  return 0;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -82,35 +96,16 @@ DonorsChooseDonationController.prototype.resourceName = 'donors-choose';
  *   Express Response object
  */
 DonorsChooseDonationController.prototype.start = function(request, response) {
-  var self = this;
```
aaronschachter: Missed this in #568 - no longer needed.

app/lib/donations/controllers/DonorsChooseDonationController.js
```diff
@@ -82,35 +96,16 @@ DonorsChooseDonationController.prototype.resourceName = 'donors-choose';
...
-
   var phone = smsHelper.getNormalizedPhone(request.body.phone);
   userModel.findOne({phone: phone}, onUserFound);
   response.send();
 
-  function onUserFound(err, doc) {
-    logger.log('debug', 'DonorsChoose.onUserFound:%s', JSON.stringify(doc));
-
-    var config;
```
aaronschachter: Same, no longer needed per `donorsChooseConfig` variable introduced in #568

PR #573, opened 2016-08-11
Donations chatbot

aaronschachter:
#### What's this PR do?
This isn't a completely 100% dev complete Science Sleuth, but it does submit donations to the DC API, and it's quite a bit of changes - so merging into`develop` now to keep future pull requests readable.  Here's what's going on:

* Refactors Donors Choose Science Sleuth donations to use a single endpoint: `/donations/donors-choose`, inspired by how [custom Slack apps use a single Action URL](https://api.slack.com/docs/message-buttons) to implement custom actions.

    * Removes now deprecated endpoints and related functions
        * `/donations/donors-choose/start`
        * `/donations/donors-choose/retrieve-firstname`
        * `/donations/donors-choose/retrieve-email`
        * `/donations/donors-choose/retrieve-zip` 

    * Removes unnecessary donation controller interface code to simplify codebase

* Refactors Donation MoCo Campaigns to use 3 OIP's: 1) Start 2) Chatbot and 3) Conversation Ended.

    * We now store the majority of conversation content into new `dcConfig` variable instead of legacy convention of storing a collection of OIP's (where we'd set conversation copy in each OIP's conversation settings, ad need to set relevant mData's, e.g. #546).  The desired `dcConfig` message is stored to the MoCo `slothbot_response` Custom Field and displayed via Liquid within our OIP conversations (refs #537)

    * `dcConfig` object will be refactored as a new `config` document (see todos section)

* Adds new helper functions for sending SMS messages:

   * `respondAndListen(member, messageText, customFields)` - opts the member into the Chatbot OIP and saves the `messageText` to the `slothbot_response` MoCo custom field. The Chatbot OIP will post back to our endpoint to continue responding to the member

    * `respond(member, messageText, customFields)` - opts the member into the Conversation Ended OIP, which does not post to an mData.

* Stores required user donation info into Mobile Commons custom fields (`email`, `first_name`, and `postal_code`) if values don't exist, instead of storing in `donation_infos` documents

    * Changes flow to prompt for all member information first. When we have all information stored, query for closest project and donate. This closes #550 and closes #567 -- if donation fails, we can start over without worrying about incomplete `donation_infos` documents and fetching an incomplete donation

* Stores count of user donations as a Mobile Commons custom field instead of the `sg_users` document corresponding to the user, removes now deprecated functions like `incrementDonationCount`. The new `ss2016_donation_count` MoCo Custom Field  is updated within `sendSuccessMessages` (which may make sense to rename as something like `respondWithSuccessConfirmation`)

[...]

PR #579, opened 2016-08-16
Saving email fix + Searching By Zip message

aaronschachter:
#### What's this PR do?
* ~~Adds support for a new `DONORSCHOOSE_CLEANSLATE_KEYWORD` that will reset a member's Donation Count if corresponding new `DONORSCHOOSE_CLEANSLATE_ENABLED` is set to `true`. It also attempts to reset the member's `first_name`, `email`, and `postal_code` but for now we'll need to manually reset those in Mobile Commons by editing the profile~~
* Adds a `msg_searching_by_zip` configuration property to inform a donor (both first time and returning) that we're finding a project within their zip code.
    * Supports MoCo Liquid style variable for postcode: A `msg_searching_by_zip` value of `'Finding a school in your zipcode {{postal_code}}...'` will render as "Finding a school in your zipcode 94117..." per calling String `replace` method in our `sendSMS` method.
* Fixes a bug saving the user's email. We confusingly also have a `email_address` custom field created in Mobile Commons, which is not the same as the default `email` field set on a profile.

* Pull down this branch and run locally
* Test the Searching By Zip message by posting to your local `/donations/donors-choose` endpoint with `phone`, `profile_first_name`, `profile_postal_code`, and `profile_email` body params set
* ~~Test the Clean Slate functionality by adding `DONORSCHOOSE_CLEANSLATE_ENABLED=true` to your .`env` file, and set a `DONORSCHOOSE_CLEANSLATE_KEYWORD`. Post the keyword locally as the `args` body param.~~

We could just reset the profile through Mobile Commons, but this aims to make testing the entire flow quicker.

#### Relevant tickets
Closes more todos in #573 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/lib/donations/controllers/DonorsChooseDonationController.js
+++ b/app/lib/donations/controllers/DonorsChooseDonationController.js
@@ -153,7 +153,7 @@ DonorsChooseDonationController.prototype.chatbot = function(request, response) {
       return;
     }
 
-    self.findProjectAndRespond(member, {email_address: firstWord});
+    self.findProjectAndRespond(member, {email: firstWord});
     return;
 
   }
--- a/app/lib/donations/controllers/DonorsChooseDonationController.js
+++ b/app/lib/donations/controllers/DonorsChooseDonationController.js
@@ -142,13 +142,15 @@ DonorsChooseDonationController.prototype.chatbot = function(request, response) {
       self.chat(member, "Whoops, that's not a valid email address. " + self.dcConfig.msg_ask_email);
       return;
     }
-    // @todo Does this ever get called?
-    self.chat(member, "Looking for a project by you, just a moment...",
-      {email_address: firstWord});
+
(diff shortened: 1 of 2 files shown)


aaronschachter:
Renamed pull request and updated description. It's not easy to reset the profile via the Mobile Commons `profile_update` endpoint: It doesn't pass key value pairs where the value is empty, and also won't save if we pass invalid postal codes or emails.

To reset a member's Donation flow:
*  Edit profile in Mobile Commons: save their first name, email, and postal code with empty values
* Scroll down to the `DONATION_COUNT_FIELDNAME` Custom Field that stores our Donation Count. For this year's run, the field name is `ss2016_donation_count`. Use the icon on the right to save the value as 0.
screen shot 2016-08-16 at 8 49 38 am
screen shot 2016-08-16 at 8 56 16 am

Issue #606, opened 2016-08-29
New Campaign bot_type


aaronschachter:
Support a new `bot_type=campaign` in chatbot endpoint to reportback for DS campaign passed by query parameter.
e.g. `/v1/chatbot?bot_type=campaign&campaign=362`

I've been debating about whether to namespace with `ds-campaign` or just use regular `campaign` (and always refer a Mobile Commons campaign as a `mobilecommons_campaign`refs #600)

Refs new work on #463 and #506.

Here's a few todos that may warrant sub-issues. Refs [DSCampaignBot spec - Google doc](https://docs.google.com/a/dosomething.org/document/d/1Exm1vHym70J7mn6ImuzsMJVm2VqHQvaGGUXh2bTFXTQ/edit?usp=sharing)

* [x] New `CampaignBotController.chatbot` function should walk the member through Campaign Signup and Campaign Reportback using a Chatbot Mobile Commons Campaign (configured in Mobile Commons with a `oip_chat`)

* [x] Northstar API integration

* [ ] "Invalid input" messages should include option for a QUESTION command. Upon texting QUESTION, member is prompted what question to ask staff. Upon member response, chatbot responds that we should get back to them soon, and member can continue in campaign
    * This could potentially be handled with a `oip_question` Opt In Path that doesn't have an mData listening. The message sent would ask the member to text back in the KEYWORD specific to the Campaign

* [ ] Support for a LATER command where member can text LATER to be reminded about completing the Campaign they've signed up for at a later date

* [x] Support for repeat reportbacks. "We've got you down for 55 nouns verbed, text back to add more"

* [ ] Easier repeat reportback. "Did you add more?" "How many more" "Wow, 300 more! Let's see a photo"

aaronschachter:
We've been creating new issues for remaining tasks that need to get in for launching CampaignBot 1.0. 

The remaining items in this issue should be opened up as separate issues if/when they're confirmed for development.

PR #627, opened 2016-09-19
Adds Northstar Client, queries for existing Northstar ID

aaronschachter:
#### What's this PR do?

Begins adding Northstar support into the CampaignBot prototype.
- Stores a `NorthstarClient` instance to `app.locals.northstarClient`
- Modifies local `users` document creation for a Northstar ID (based on the Mobile Commons `northstar_id` Profile Field) to query Northstar to store our User's First Name, Email, and Phoenix ID 
- Run `npm install` to add Northstar JS
- Delete your `users` collection locally, and then test posting to the CampaignBot with your northstar id set as a `profile_northstar_id`. Verify a user document is created with your mobile, email, first name, and Phoenix ID.
- Storing Phoenix ID will make it easy to post Signups and Reportbacks to Phoenix for now, since Northstar POST Signup/Reportback doesn't support a `user` input parameter (https://github.com/DoSomething/northstar/issues/429)
- TODO: Pin package.json to new Northstar JS release (or just always leave as breaking for now until development is finished) 
- TODO: If we don't have an incoming `northstar_id`, check to see if Northstar User exists for mobile (which it probably won't). If not, create new Northstar User and corresponding Gambit User doc
- TODO We'll also want use the `northstarClient` to query for the current User's Signups for a given campaign, using [`app.locals.northstarClient.Signups.index`](https://github.com/DoSomething/northstar-js/blob/b4a7b985b6681ddac38e82dc53019b65c6488574/lib/northstar-endpoint-signups.js#L33)
#### Relevant tickets
#606 - begins Northstar integration
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -104,27 +104,24 @@ CampaignBotController.prototype.chatbot = function(req, res) {
 CampaignBotController.prototype.createUserAndPostSignup = function(req, res) {
   var self = this;
 
-  app.locals.northstarClient.Users.get('id', req.user_id).then(function(user) {
-
-    dbUsers.create({
-
-      _id: req.user_id,
-      mobile: req.user_mobile,
-      first_name: user.firstName,
-      email: user.email,
-      phoenix_id: user.drupalID,
-      campaigns: {}
-
-    }).then(function (userDoc) {
-
+  app.locals.northstarClient.Users.get('id', req.user_id)
+    .then(function(user) {
+      const userFields = {
+        _id: req.user_id,
+        mobile: req.user_mobile,
+        first_name: user.firstName,
+        email: user.email,
+        phoenix_id: user.drupalID,
+        campaigns: {} ,   
+      };
+      return dbUsers.create(userFields);
+    })
+    .then(function (userDoc) {
       self.user = userDoc;
       logger.debug('%s created user', self.loggerPrefix(req));
       self.postSignup(req, res);
-
-    });
-
-  }, function(err) {self.handleError(req, res, err)});
-
+    })
+    .catch(function(err) {self.handleError(req, res, err)});
 }
 
 /**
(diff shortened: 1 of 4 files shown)


aaronschachter:
Added `DS_REST_API_KEY` and `DS_REST_API_BASEURI` and verified User doc created in staging.

api/controllers/CampaignBotController.js
```diff
@@ -104,24 +104,27 @@ CampaignBotController.prototype.chatbot = function(req, res) {
 CampaignBotController.prototype.createUserAndPostSignup = function(req, res) {
   var self = this;
 
-  var newUser = {
-    _id: req.user_id,
-    mobile: req.user_mobile,
-    campaigns: {}
-  };
+  app.locals.northstarClient.Users.get('id', req.user_id).then(function(user) {
```
sergii-tkachenko: Just a general suggestion: instead of thening `dbUsers.create()` model inside `Users.get.then()`, you can instead return from it `dbUsers.create()` and `then()` the result one level above. Thus both thens are handled by the `catch` at the end. Something like this:

```js

app.locals.northstarClient.Users.get('id', req.user_id)
  .then(function(user) {
    const userFields = {
      _id: req.user_id,
      mobile: req.user_mobile,
      first_name: user.firstName,
      email: user.email,
      phoenix_id: user.drupalID,
      campaigns: {},
    };
    return dbUsers.create(userFields);
  }
  .then(function (userDoc) {
    self.user = userDoc;
    logger.debug('%s created user', self.loggerPrefix(req));
    self.postSignup(req, res);

  })
  .catch(function(err) {self.handleError(req, res, err)});
```

sergiitk (approved):
👍

api/controllers/CampaignBotController.js
aaronschachter: That's way cleaner, thanks @sergii-tkachenko ! I'll make that change.

Issue #549, opened 2016-08-03
Handle connection errors to Phoenix


aaronschachter:
Bigger picture issue to help stop #506. If Phoenix goes down and a member attempts to submit a reportback (which happens after completing the flow by submitting `why_participated`) -- the reportback is stuck in the `reportbacks` collection and currently can only be saved by a manual import (#544) 

Ideally our Reportback service could implement a queue to handle saving Reportbacks, and our end users submit Reportbacks without getting connection errors back.

On mobile app, we inform the user with an error message that something wrong has occurred. The user must then remember to go back to the campaign later and try re-uploading the Reportback.

On SMS, the user doesn't even get a message back -- there's just no response. A band-aid option could be to at least respond to the member back with a "Oops please try again later" but we'd still need code to handle coming back to an incomplete campaign (issue upcoming)

aaronschachter:
Taking a look at the `node-request-retry` package that I didn't realize is a DoSomething project. Just nothing that we stop re-requesting failed Phoenix post after 3 times, [per the default](https://github.com/DoSomething/node-request-retry/blob/master/index.js#L5).

aaronschachter:
The ideal would be adding support into Quicksilver API (https://github.com/DoSomething/quicksilver-api/issues/98) to queue our Registration, Signup or Reportback POST requests. Upon a 200 OK back, we send the next message to our User to continue in flow without relying on the returned ID.

aaronschachter:
Marking this as resolved for now -- until we get a Quicksilver queue for Signup and Reportbacks in place (https://github.com/DoSomething/quicksilver-api/issues/98). CampaignBot will return a 500 if a Phoenix request fails. If the member was in the middle of a reportback, texting the keyword in again will retry the reportback request.

Improvements could be to throw a 404 and handle accordingly "Oops, our site is down now.. try again in a bit", but let's create a new issue if that's something we want to add.

PR #629, opened 2016-09-23
Adds Phoenix JS, posts Campaign Signup to Phoenix

aaronschachter:
#### What's this PR do?
- Adds `phoenix-js` - pulls from `master` branch for now 🔥 
- Adds an authorized `phoenixClient` to our `app.locals` (refs #627)
- Modifies `/campaigns/models/Signup.js` to store `_id` as Number instead of an ObjectId
- Posts Signup to Phoenix and creates a document in the Signups collection with newly inserted Signup ID
- Delete any existing `signups` documents or `user.campaigns` properties -- you'll get a Cast error
- Post to CampaignBot for a campaign you're not signed up for
- To remove signup, you'll need to remove the Signup from Phoenix using the webform (logged in as admin), but will also need to modify the `user.campaigns` property in your Gambit `users` collection to clear out the Signup stored for the Campaign.

I've been debating on whether we should write the Signup anyway, regardless of whether the Phoenix POST succeeds or fails. Returning an error (instead of keeping track of our own Signup ObjectID) seemed like the simpler approach, but keeping our own cache also has its benefits if Phoenix is down 🚑 
#### Relevant tickets
#606, #549 https://github.com/DoSomething/quicksilver-api/issues/98
#### Checklist
- [x] Set new environment variables (`DS_NORTHSTAR_API_BASEURI`, `DS_NORTHSTAR_API_KEY`,`DS_PHOENIX_API_BASEURI`,  `DS_PHOENIX_API_USERNAME`, `DS_PHOENIX_API_PASSWORD`, `DS_API_POST_SOURCE`)
- [x] Test on staging.

### DIFF
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -404,18 +404,29 @@ CampaignBotController.prototype.supportsMMS = function(req, res) {
 CampaignBotController.prototype.postSignup = function(req, res) {
   this.debug(req, 'postSignup');
 
-  // @todo Query DS API to check if Signup exists, post to API to get _id if not
+  // @todo Sanity check by using Northstar getSignups for user/campaign before
+  // requesting POST signup from Phoenix
 
-  return dbSignups
-    .create({
-      campaign: this.campaignId,
-      user: req.user_id,
+  const postData = {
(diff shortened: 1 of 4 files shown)


sergiitk (commented):
Small note.

PR #628, opened 2016-09-22
Updates Mongoose and replaces callbacks with promises

aaronschachter:
#### What's this PR do?
- Updates Mongoose to 4.6.1
- Replaces `findOne` calls with `findById`
- Switches `create`, `findById`, `save` callbacks with promises [ref](http://eddywashere.com/blog/switching-out-callbacks-with-promises-in-mongoose/)
- Abstracts the `dbUsers.findOne` call into a `loadUserAndSignup` function
- Renames `continueReportbackSubmission` as `loadReportbackSubmission`
- Renames `startReportbackSubmission` as `createReportbackSubmission`
- Removes `self = this` noise
#### How should this be reviewed?

Test CampaignBot locally, on staging upon deploy.
#### Any background context you want to provide?

Next up, adding relevant Phoenix JS and Northstar JS networking calls.
#### Checklist
- [x] Tested on staging.

PR #631, opened 2016-09-24
CampaignBotController Class

aaronschachter:
#### What's this PR do?
Looks like a lot of changes, but just refactors `CampaignBotController` to use `class` instead of `prototype`. 
* Refactors `chatbot(req, res)` as `constructor(req, res)`
* Defines non-constructor functions in alphabetical order to make things easier to find

Verify CampaignBot still works.

Looking to make similar changes to the `DonorsChooseController`, and eventually add our DS [ESLint config](https://github.com/DoSomething/eslint-config).

#### Relevant tickets
https://github.com/DoSomething/gambit/pull/628#pullrequestreview-1381427

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -1,548 +1,457 @@
 "use strict";
 
+/**
+ * Imports.
+ */
 const logger = rootRequire('lib/logger');
 const mobilecommons = rootRequire('lib/mobilecommons');
 const phoenix = rootRequire('lib/phoenix')();
 const helpers = rootRequire('lib/helpers');
-
 const connOps = rootRequire('config/connectionOperations');
 const dbRbSubmissions = require('../models/campaign/ReportbackSubmission')(connOps);
 const dbSignups = require('../models/campaign/Signup')(connOps);
 const dbUsers = require('../models/User')(connOps);
 
+/**
+ * Setup.
+ */
 const CMD_REPORTBACK = (process.env.GAMBIT_CMD_REPORTBACK || 'P');
 
 /**
- * CampaignBotController
- * @constructor
- * @param {integer} campaignId - our DS Campaign ID
+ * CampaignBotController.
  */
-function CampaignBotController(campaignId) {
-  this.campaignId = campaignId;
-  this.campaign = app.getConfig(app.ConfigName.CAMPAIGNS, campaignId);
-  if (!this.campaign) {
-    return;
-  }
+class CampaignBotController {
 
-  // @todo Config variable for default CampaignBot ID
-  // @todo Store campaignbot property on our config.campaigns to override bot
-  // on a per DS Campaign basis.
-  this.bot = app.getConfig(app.ConfigName.CAMPAIGNBOTS, 41);
-
-  var mobileCommonsCampaign = this.campaign.staging_mobilecommons_campaign;
-
-  if (process.env.NODE_ENV === 'production') {
-    mobileCommonsCampaign = this.campaign.current_mobilecommons_campaign;
-  }
-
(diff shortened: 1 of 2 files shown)


api/controllers/CampaignBotController.js
```diff
@@ -1,548 +1,454 @@
...
-  }
-
-  // @todo Config variable for default CampaignBot ID
-  // @todo Store campaignbot property on our config.campaigns to override bot
-  // on a per DS Campaign basis.
-  this.bot = app.getConfig(app.ConfigName.CAMPAIGNBOTS, 41);
+class CampaignBotController {
 
-  var mobileCommonsCampaign = this.campaign.staging_mobilecommons_campaign;
+  constructor(req, res) {
```
aaronschachter: @sergii-tkachenko @deadlybutter This is related to [whether to define our current User on our `CampaignBotController` or in Express request `req`]( https://github.com/DoSomething/gambit/pull/628#pullrequestreview-1236088). Any opinion on which is cleaner -- creating`this.req` and `this.res` properties on the `CampaignBotController` vs passing them to every function? 

e.g.  `this.loadUserAndSignup()` instead of `this.loadUserAndSignup(req, res)`

deadlybutter: In the current setup I'd prefer `this.loadUserAndSignup`, otherwise you'll have problems dealing with multiple requests. A more overarching solution would be making a promise chain & passing the user in that, keeping `req` & `res` within your route function.

deadlybutter: So..

```
controller.loadUser(email).then(controller.signup).then(() => { res.status(200) });

aaronschachter: Thanks @deadlybutter -- why would there be problems with multiple requests though? We create a new controller in each request in[ `router.js`](https://github.com/DoSomething/gambit/blob/facd755741e43f4051150efb84a4ff3841c80abe/api/router.js#L44)

deadlybutter: In that case you're ok! The approach sounds scary though, rebuilding + destroying a new controller for every request sounds expensive vs reusing a single controller

aaronschachter: That's good to know for sure...  I've been following examples from [how the legacy code's been doing it](https://github.com/DoSomething/gambit/blob/1.0.0/app/lib/sms-games/index.js), I can def having less objects to build/destroy would be a plus because we need to do it for Current User, Campaign, Signup, Reportback in each CampaignBot request.  

Your code above https://github.com/DoSomething/gambit/pull/631#discussion_r80524106 looks very clean, I see what you mean about keeping `req` and `res` limited to `router.js` only ... or potentially [something like a policies directory.](https://github.com/DoSomething/quicksilver-api/blob/master/api/policies/sessionAuth.js)

deadlybutter: Yeah in theory that line of code could actually work! I could be wrong here, but the policies directory seems to be for managing your request session, which seems tricker than just using promises

aaronschachter: Indeed. moving all of the routing logic like
- if we don't have a cached user, create a user, then check for cached signup, if no signup check Phoenix for a Signup, if no Signup, post to Phoenix and cache new signup
- if we have a user, signup, and draft reportback submission, and don't have quantity yet, ask for quantity
- etc

to someplace like `router.js` def makes sense, and our `CampaignBotController` just has functions that create the cache documents, get/post to API etc.

Solid JS'ing @deadlybutter -- thanks for the feedback!

deadlybutter: image

deadlybutter: giphy

aaronschachter: #632

Issue #636, opened 2016-09-28
CampaignBot: create Northstar User if doesn't exist


aaronschachter:
There's a `northstar_id` Custom Field that some Mobile Commons profiles will have, but not guaranteed for all.

For an incoming CampaignBot request without a Northstar ID:

1. Query Northstar to see if we have a Northstar User for the given mobile number. If we do:

    * [x] Create a User model locally in Gambit

    * [x] Update member profile `northstar_id` Custom Field in Mobile Commons

2. If no User exists:

    * [x] Create User in Northstar on behalf of member, and save locally in Gambit

    * [x] Update member profile `northstar_id` Custom Field  in Mobile Commons

PR #637, opened 2016-09-29
Adds staff-only command to clear Signup cache

### LINKED ISSUE #634: Add staff-only command to clear Signup cache
Texting in a staff-only command should clearing out your user's Signups cache in Gambit. Needed for testing and re-re-testing Signup flow.

Only allow for users where `role` is set to `'admin'` or `'staff'` -- blocked by https://github.com/DoSomething/northstar-js/issues/26

The steps for re-testing a Campaign XYZ Signup with your user:

1. Remove Signup for Campaign XYZ via Phoenix admin UI
2. Text in the Clear Cache keyword to Campaign XYZ to reset Signup cache

aaronschachter:
#### What's this PR do?
- Checks for whether current User has `role` set to `'staff'` or `'admin`'  and has texted in a Clear Cache command. If true, delete the cached Signup model and call `getCurrentSignup(req)` 

Test by locally modifying your User's `role` property in the `users` collection, and text in clear cache keyword (set by `GAMBIT_CMD_CLEAR_CACHE` environment variable). 
- Text Clear Cache keyword when a Signup exists in Phoenix, confirm existing Signup is cached to Gambit Signups collection
- Remove signup from Phoenix admin UI, text Clear Cache keyword and confirm new Signup ID is created in Phoenix and cached to Signups collection

For internal testing of Signup flow to avoid API error code 500 returned when resubmitting a Signup for a a given Campaign and User.
#### Relevant tickets
- Fixes #634
- https://github.com/DoSomething/northstar-js/pull/27
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -254,8 +254,9 @@ class CampaignBotController {
     }
 
     const input = helpers.getFirstWord(req.incoming_message);
+    const isCommand = input && input.toUpperCase() === cmdClear.toUpperCase();
 
-    return ( input && input.toUpperCase() === cmdClear.toUpperCase() );
+    return isCommand;
   }
 
   /**
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -225,6 +225,27 @@ class CampaignBotController {
   }
 
   /**
+   * Returns whether incoming request is an authorized cache clear command.
+   * @param {object} req
+   * @return {bool}
+   */
+  isCommandClearCache(req) {
+    const cmdClear =  process.env.GAMBIT_CMD_CLEAR_CACHE;
+
+    if (!cmdClear) return false;
+
+    if (req.user.role !== 'staff' && req.user.role !== 'admin') {
+      logger.warn(`${this.loggerPrefix(req)} unauthorized clear cache`);
+
+      return false;
+    }
+
+    const input = helpers.getFirstWord(req.incoming_message);
+
+    return ( input && input.toUpperCase() === cmdClear.toUpperCase() );
+  }
+
+  /**
    * Returns whether incomingMessage should begin a Reportback conversation.
    * @param {string} incomingMessage
    * @return {bool}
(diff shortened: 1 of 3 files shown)


api/controllers/CampaignBotController.js
```diff
@@ -241,6 +264,20 @@ class CampaignBotController {
    * @return {object}
    */
   loadCurrentSignup(req, id) {
+    this.debug(req, `loadCurrentSignup:${id}`);
+
+    if (this.isCommandClearCache(req)) {
+      this.debug(req, `isCommandClearCache`);
+
+      return app.locals.db.signups
+        .findByIdAndRemove(id)
```
aaronschachter: Not needed for merge, but we'll eventually want support for clearing your cached User model too (refs #636) -- I'm thinking a better approach may be to remove all Signup documents cached in the User's `campaigns` property upon deleting the User document -- clearing out the entire User cache instead of just the User's cache for this Campaign. That way we won't run into any Mongo errors of trying to create Signup documents with duplicate Signup id's.

api/router.js
```diff
@@ -101,10 +103,7 @@ function campaignBot(req, res) {
       req.user = user;
       const signupId = user.campaigns[req.campaign_id];
 
-      if (signupId) {
-        controller.debug(req, `loadSignup:${signupId}`);
-        return controller.loadCurrentSignup(req, signupId);
-      }
+      if (signupId) return controller.loadCurrentSignup(req, signupId);
```
sergii-tkachenko: Please use curly brackets here.

api/controllers/CampaignBotController.js
```diff
@@ -225,13 +227,36 @@ class CampaignBotController {
...
+
+    if (req.user.role !== 'staff' && req.user.role !== 'admin') {
+      logger.warn(`${this.loggerPrefix(req)} unauthorized clear cache`);
+
+      return false;
+    }
+
+    const input = helpers.getFirstWord(req.incoming_message);
+
+    return ( input && input.toUpperCase() === cmdClear.toUpperCase() );
```
sergii-tkachenko: Extra parentheses. Move the expression to own line and assign to a var so it's easier to debug.

api/router.js
aaronschachter: I wasn't adding per https://github.com/airbnb/javascript#blocks--braces... 
````
// good
if (test) return false;

// bad
function foo() { return false; }
````

Looking to add #638 next to avoid syntax changes in future PR's.

api/controllers/CampaignBotController.js
aaronschachter: Uses https://github.com/DoSomething/gambit/pull/637/commits/c2e055111897ee53283962589072af8469619aec instead of deleting the Signup model.

If user texts in the clear cache command, it will reset their `user.campaigns` cache for all Campaigns (not just the current Campaign they're texting to).

sergiitk (approved):
👍

aaronschachter:
Added documentation to https://github.com/DoSomething/gambit/wiki/Development#testing

PR #647, opened 2016-10-03
Adds User.current_campaign property, CampaignBot MobileCommons OIP config variable

### LINKED ISSUE #633: Refactor CampaignBot to use single Mobile Commons Campaign
Current build requires creating a new Mobile Commons Chatbot campaign per each DS Campaig. Refactor to use the opt-in path for a single default `MobileCommonsChatbotCampaign` model.

This may start getting into handling signups differently, and thinking through the various routes for a DS Campaign signup:

* Text in a keyword and respond back to confirm signup for Campaign
* Respond to a SMS broadcast with affirmative to signup for Campaign
* Signup via web, mobile apps

aaronschachter:
#### What's this PR do?
- Changes all CampaignBot Mobile Commons `profile_update` requests to post to a single Opt-in Path, defined by `CAMPAIGNBOT_MOBILECOMMONS_OIP` config variable (set to 214465 for staging)
  - Removes deprecated `current_mobilecommons_campaign` and `staging_mobilecommons_campaign` properties from our Campaign `config` model
- Adds a `User.current_campaign` property to store the DS Campaign the User is currently participating in:
  - If a `campaign=123` is passed in the query string, use it to load the DS Campaign 123 and saving User activity to. Upon creating or loading signup, set the user's `current_campaign`property  to 123.
  - If a `campaign` is not passed as a query parameter, load the DS Campaign saved to our `user.current_campaign` property
- Sets a `CAMPAIGNBOT_ID` config variable to loading our default CampaignBot message copy (was previously hardcoded to id [41](http://dev-gambit-jr.pantheonsite.io/wp-json/wp/v2/campaignbots/41))
- Deprecates the `start=true` parameter in favor of checking for the `campaign` parameter -- only time we should pass a `campaign` parameter is to change current chatbot conversation to signup/continue confirmation for the given campaign.

Test on staging. I made following changes in Mobile Commons to implement:
- Our OIP 214465 triggers our "Staging CampaignBot" mData which posts to `/v1/chatbot` -- no `campaign` is passed into the query string, so all responses to this OIP will load the saved User's `current_campaign`
- Our Bumble Bands keyword / mData posts to `/v1/chatbot?campaign=2299`
- Our Two Books keyword / mData posts to `/v1/chatbot?campaign=2070`

We'll want to cache a DS Campaign's `status` property soon to determine whether the Campaign should be available to participate in via SMS.
#### Relevant tickets

Fixes #633
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/config/locals.js
+++ b/config/locals.js
@@ -125,10 +125,8 @@ module.exports.load = function () {
 
   return Promise.all(locals).then(() => {
     const CampaignBotController = rootRequire('api/controllers/CampaignBotController');
-    // TODO: We'll need to loop through all campaignBots and store as app.locals.campignBots map.
-    // For now we only have one default CampaignBot (id 41) to use for all DS Campaigns.
-    // TODO: Add config variable to avoid hardcoding default CampaignBot.
-    const campaignBot = app.locals.configs.campaignBots[41];
+    // TODO: Support multiple campaignBots, to allow campaigns to override default message copy.
+    const campaignBot = app.locals.configs.campaignBots[process.env.CAMPAIGNBOT_ID];
     app.locals.campaignBot = new CampaignBotController(campaignBot);
 
     const DonorsChooseBotController = rootRequire('api/controllers/DonorsChooseBotController');
--- a/config/router-chatbot.js
+++ b/config/router-chatbot.js
@@ -40,15 +40,6 @@ router.post('/', (req, res) => {
   const campaign = app.locals.configs.campaigns[req.campaign_id];
   req.campaign = campaign; // eslint-disable-line no-param-reassign
 
-  // TODO: Mobile Commons Campaign will be shared by all DS Campaigns
-  // @see https://github.com/DoSomething/gambit/issues/633
-  let mocoId = req.campaign.staging_mobilecommons_campaign;
-  if (process.env.NODE_ENV === 'production') {
-    mocoId = req.campaign.current_mobilecommons_campaign;
-  }
-  const mobilecommonsCampaign = app.locals.configs.chatbotMobileCommonsCampaigns[mocoId];
-  const mobilecommonsOptinPath = mobilecommonsCampaign.oip_chat;
-
   return controller
     .loadUser(req.user_id)
     .then(user => {
@@ -91,7 +82,7 @@ router.post('/', (req, res) => {
     })
     .then(msg => {
       controller.debug(req, `sendMessage:${msg}`);
-      mobilecommons.chatbot(req.body, mobilecommonsOptinPath, msg);
+      mobilecommons.chatbot(req.body, process.env.CAMPAIGNBOT_MOBILECOMMONS_OIP, msg);
 
       return res.send({ message: msg });
     })
--- a/config/router-chatbot.js
+++ b/config/router-chatbot.js
@@ -17,6 +17,8 @@ router.post('/', (req, res) => {
   req.incoming_image_url = req.body.mms_image_url;
   /* eslint-enable no-param-reassign */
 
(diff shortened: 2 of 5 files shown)


aaronschachter:
This branch as is will sign the user up for the campaign upon texting in the keyword -- without a confirmation. I think it makes sense to ask for a confirmation before posting Signup -- only post the Campaign Signup if the user texts back an affirmative response to the initial message they receive upon the mData's POST our `chatbot?start=true&campaign=X`

aaronschachter:
On second thought, I think we're ok to hold off for now on adding asking for and handling a Signup confirmation. It adds complexity to our `chatbot-router` logic and we don't do this currently with legacy campaigns. 

I do think it make sense to drop the use of the `start` parameter, we can just check for the existence of a `campaign` parameter instead in [`collectReportbackQuantity`:
](https://github.com/DoSomething/gambit/blob/ff394419e463c543b56eebebabbca0262b25cccb/api/controllers/CampaignBotController.js#L39)
````
// If a campaign has been passed -- we're beginning a new conversation for the given campaign.
//  Send message asking for property (disregarding whatever text member has sent -- in this case the start KEYWORD)
if (req.query.campaign || ask) {
  return this.renderResponseMessage(req, `ask_${property}`);
}
````

File: config/router-chatbot.js [hunk 1]
```diff
@@ -91,7 +88,9 @@ router.post('/', (req, res) => {
     })
     .then(msg => {
       controller.debug(req, `sendMessage:${msg}`);
-      mobilecommons.chatbot(req.body, mobilecommonsOptinPath, msg);
+      controller.setCurrentCampaign(req.user, req.campaign_id);
+
+      mobilecommons.chatbot(req.body, process.env.CAMPAIGNBOT_MOBILECOMMONS_OIP, msg);
```
sergii-tkachenko: Is there a default value for the env var?

aaronschachter: I can hardcode it back to 41...

config/locals.js
```diff
@@ -125,10 +125,8 @@ module.exports.load = function () {
 
   return Promise.all(locals).then(() => {
     const CampaignBotController = rootRequire('api/controllers/CampaignBotController');
-    // TODO: We'll need to loop through all campaignBots and store as app.locals.campignBots map.
-    // For now we only have one default CampaignBot (id 41) to use for all DS Campaigns.
-    // TODO: Add config variable to avoid hardcoding default CampaignBot.
-    const campaignBot = app.locals.configs.campaignBots[41];
+    // TODO: Support multiple campaignBots, to allow campaigns to override default message copy.
+    const campaignBot = app.locals.configs.campaignBots[process.env.CAMPAIGNBOT_ID];
```
sergii-tkachenko: Also, here?

sergii-tkachenko:
Looks good, just small question.

config/router-chatbot.js
sergii-tkachenko: Ah, ok. We just need to be sure it'll throw some kind of error if the value is undefined.

sergiitk (approved):
👍

config/router-chatbot.js
aaronschachter: Good point -- I'll add a check. Thanks @sergii-tkachenko !

aaronschachter:
Adding screenshots here to link to from newly created [Chatbot wiki page](https://github.com/DoSomething/gambit/wiki/Chatbot):

## chatbot Mobile Commons OIP
screen shot 2016-10-04 at 10 01 14 am

## default chatbot mData:
screen shot 2016-10-04 at 10 11 45 am

## creating Campaign Signup keyword:
screen shot 2016-10-04 at 10 12 09 am

PR #653, opened 2016-10-06
Adds CAMPAIGNBOT_CAMPAIGNS config variable, Gambit Jr. support for Campaign tagline + reportback confirmation message 

aaronschachter:
#### What's this PR do?

`config/locals.js`
- On app start, loads `app.locals.campaigns` from Phoenix JS for id's defined by new config variable `CAMPAIGNBOT_CAMPAIGNS`.  Expects a comma separated list, e.g. `CAMPAIGNBOT_CAMPAIGNS=362,2900,2299`
- Refactors `load()` to create our `app.locals.clients` and then return all other promises; moves creating `app.locals.controllers` into separate `loadDonorsChooseBotController(:id)`, `loadCampaignBotControllers()`, `loadSlothBotController()` functions that return promises.
- Load Gambit Jr. bots on app start when we create our controllers, deprecating the need for the `chatbot/sync` endpoint (which would still require an app restart anyway upon posting, since our configs are loaded in memory in `locals.loadConfigs` -- not by querying Mongo) 
- Removes the `api/config` models (`Campaign`, `CampaignBot`, `DonorsChooseBot`) for now, because we're loading these objects via `app.locals.campaigns` and `app.locals.controllers` on app start (and not reading from them in case of failure). 
  - I'd prefer to add these models back in if/when we add functionality to read from DB in case of failure to connect to Phoenix or Gambit Jr (if this is something we need - let's create an issue). It's convenient not keeping a cache of these models for now while we're still in development (e.g. we didn't need to make model changes to add new `Campaign` properties)

`app/controllers/CampaignBotController.js`
- Makes the Campaign tagline and reportback confirmation message available as Gambit Jr. template variables `{{tagline}}` and `{{rb_confirmed}}` - refs https://github.com/DoSomething/phoenix-js/pull/14

`lib/gambit-junior.js`
- Refactors the `gambit-junior` helper file that returns parsed `donorschoosebots` and `campaignbots` from Gambit Jr. API to use superagent instead of request
- Added  to `eslint` in `package.json`

👀 
#### Checklist
- [x] Set `CAMPAIGNBOT_CAMPAIGNS` config variable on staging and production
- [x] Documentation added for new features/changed endpoints. -- https://github.com/DoSomething/gambit/wiki/Gambit-Jr.
- [x] Tested on staging.


sergiitk (approved):
👍

aaronschachter:
This branch is failing on Heroku staging with  `uncaughtException: app.locals.clients.phoenix.Campaigns.index is not a function`, which was introduced a day ago in https://github.com/DoSomething/phoenix-js/pull/14 😕

aaronschachter:
Fixed deployment by setting `NODE_MODULES_CACHE` [config variable](https://devcenter.heroku.com/articles/nodejs-support#cache-behavior) to `false` and redeploying.

PR #654, opened 2016-10-06
AgentView support

aaronschachter:
#### What's this PR do?
- Namespaces Mobile Commons specific config variables with a `MOBILECOMMONS_` prefix:
  - `MOBILECOMMONS_OIP_CAMPAIGNBOT`, `MOBILECOMMONS_OIP_DONORSCHOOSEBOT`, `MOBILECOMMONS_OIP_DONORSCHOOSEBOT_ENDED`, `MOBILECOMMONS_OIP_DONORSCHOOSEBOT_ERROR`
- Adds a new `GAMBIT_CMD_MEMBER_SUPPORT` config variable. If the member texts in the value during a CampaignBot conversation, they are opted into a Mobile Commons OIP defined by new config variable `MOBILECOMMONS_OIP_AGENTVIEW`.
  - The AgentView OIP conversation should be set to **NOT** trigger any mData's, which means all subsequent messages the member sends will be added to the Mobile Commons Agent View queue for member support to respond to, until they text in another Mobile Commons keyword.
  - Sends new CampaignBot `member_support` message to the member when they are opted into the Agent View OIP. This message should contain the `{{keyword_continue}}` variable to inform the member how to continue in CampaignBot after they have submitted their support questions.  `{{keyword_continue}}` renders our new `MOBILECOMMONS_KEYWORD_CAMPAIGNBOT`config variable.
  - The `MOBILECOMMONS_KEYWORD_CAMPAIGNBOT` should be set to the keyword defined on our CampaignBot mData, which posts to chatbot without a `campaign` (the same mData that our `MOBILECOMMONS_OIP_CAMPAIGNBOT` posts to). This allows the member to text in a keyword to continue where they left (loading their saved `current_campaign`).
  - The end result of all this makes our `member_support` copy look something like this: 
    
    > Text back your question and I'll do my best to answer back within 24 hours. If you'd like to continue on {{title}}, text back {{keyword_continue}}.
- Adds additional Gambit Jr. support for `{{cmd_reportback}}` and `{{cmd_member_support}}` for Gambit commands to use in CampaignBot messages.  Here's what an example `menu_completed` could read like: 
  
  > Got you down for {{quantity}} {{rb_noun}} {{rb_verb}}. If you have {{rb_verb}} more, text {{cmd_reportback}}. If you have a question, text {{cmd_member_support}}
- Refactors `isCommandCacheClear(req)` and `isCommandReportback(req)` to single `isCommand(req, type)` function with expected `type` values of `cache_clear`, `reportback`, and `member_support`

👀 
#### Relevant tickets
#606
#### Checklist
- [x] Added required config vars on staging 
- [x] Tested on staging.
- [ ] Added required config vars on production

### DIFF
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -254,7 +254,7 @@ class CampaignBotController {
 
     const result = this.parseCommand(req) === configValue.toUpperCase();
 
-    if (type === 'clear_cache') {
+    if (result && type === 'clear_cache') {
       if (!this.isStaff(req.user)) {
         logger.warn(`${this.loggerPrefix(req)} unauthorized command clear_cache`);
 
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -247,7 +247,7 @@ class CampaignBotController {
     const configName = `GAMBIT_CMD_${type.toUpperCase()}`;
     const configValue = process.env[configName];
     if (!configValue) {
-      logger.error(`${this.loggerPrefix(req)} process.env.${configName} is undefined`);
+      logger.warn(`${this.loggerPrefix(req)} process.env.${configName} is undefined`);
 
       return false;
     }
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -438,13 +438,13 @@ class CampaignBotController {
 
     msg = msg.replace(/{{br}}/gi, '\n');
     msg = msg.replace(/{{title}}/gi, campaign.title);
-    msg = msg.replace(/{{tagline}}/gi, campaign.tagline);
+    msg = msg.replace(/{{tagline}}/i, campaign.tagline);
     msg = msg.replace(/{{rb_noun}}/gi, campaign.reportbackInfo.noun);
     msg = msg.replace(/{{rb_verb}}/gi, campaign.reportbackInfo.verb);
-    msg = msg.replace(/{{rb_confirmed}}/gi, campaign.reportbackInfo.confirmationMessage);
-    msg = msg.replace(/{{cmd_reportback}}/gi, process.env.GAMBIT_CMD_REPORTBACK);
-    msg = msg.replace(/{{cmd_member_support}}/gi, process.env.GAMBIT_CMD_MEMBER_SUPPORT);
-    msg = msg.replace(/{{keyword}}/gi, campaign.keywords);
+    msg = msg.replace(/{{rb_confirmed}}/i, campaign.reportbackInfo.confirmationMessage);
+    msg = msg.replace(/{{cmd_reportback}}/i, process.env.GAMBIT_CMD_REPORTBACK);
+    msg = msg.replace(/{{cmd_member_support}}/i, process.env.GAMBIT_CMD_MEMBER_SUPPORT);
+    msg = msg.replace(/{{keyword}}/i, process.env.MOBILECOMMONS_KEYWORD_CAMPAIGNBOT);
 
     if (req.signup) {
       let quantity = req.signup.total_quantity_submitted;
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -31,7 +31,7 @@ class CampaignBotController {
   collectReportbackProperty(req, property, ask) {
     this.debug(req, `collectReportbackProperty:${property}`);
 
-    if (req.query.campaign || ask) {
+    if (req.query.campaign || ask || this.isKeywordCampaignBot(req)) {
       return this.renderResponseMessage(req, `ask_${property}`);
     }
 
@@ -267,6 +267,18 @@ class CampaignBotController {
   }
 
   /**
+   * Returns whether incoming request is the CampaignBot Mobile Commons keyword.
+   * @param {object} req
+   * @return {bool}
+   */
+  isKeywordCampaignBot(req) {
+    const keyword = process.env.MOBILECOMMONS_KEYWORD_CAMPAIGNBOT;
+    const isKeyword = this.parseCommand(req) == keyword.toUpperCase();
+
+    return isKeyword;
+  }
+
+  /**
    * Loads Signup model if exists for given id, else get/create from API.
    * @param {object} req - Express request
    * @param {number} id - DS Signup ID
@@ -320,7 +332,7 @@ class CampaignBotController {
 
   /**
    * Parse incoming request as Gambit command.
-  */
+   */
   parseCommand(req) {
     return helpers.getFirstWordUppercase(req.incoming_message);
   }
@@ -444,7 +456,8 @@ class CampaignBotController {
       msg = msg.replace(/{{quantity}}/gi, quantity);
     }
 
-    if (req.query.campaign && req.signup && req.signup.draft_reportback_submission) {
+    const revisiting = req.query.campaign && req.signup && req.signup.draft_reportback_submission;
+    if (revisiting || this.isKeywordCampaignBot(req)) {
       // TODO: New bot property for continue draft message
       const continueMsg = 'Picking up where you left off on';
       msg = `${continueMsg} ${campaign.title}...\n\n${msg}`;
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -31,7 +31,7 @@ class CampaignBotController {
   collectReportbackProperty(req, property, ask) {
     this.debug(req, `collectReportbackProperty:${property}`);
 
-    if (req.query.campaign || ask || this.isKeywordCampaignBot(req)) {
+    if (ask || req.body.keyword) {
       return this.renderResponseMessage(req, `ask_${property}`);
     }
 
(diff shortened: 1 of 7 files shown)


aaronschachter:
Latest commits have been working well on staging, until Mobile Commons requests back to my phone display as failed. Don't see any errors in our Gambit app logs though 😕 

**Note** I had to alter our chatbot Mobile Commons OIP to force sending MMS today in order to avoid rendered message content truncating at 160 chars -- wonder if that could be the issue?

screen shot 2016-10-06 at 9 57 21 pm

aaronschachter:
Following up on the last comment, failed messages seem related to my profile specifically and due to a Mobile Commons "aggregator problem" with my phone and MMS messages (not SMS messages).

I did change our CampaignBot Staging OIP to "Force MMS" for the time being while we're in development -- this change was made to avoid our Gambit chatbot copy getting truncated (and put off on engineering for non-MMS members until we have our prototype finished).

api/controllers/CampaignBotController.js
```diff
@@ -36,7 +31,7 @@ class CampaignBotController {
   collectReportbackProperty(req, property, ask) {
     this.debug(req, `collectReportbackProperty:${property}`);
 
-    if (req.query.campaign || ask) {
+    if (req.query.campaign || ask || this.isKeywordCampaignBot(req)) {
```
sergii-tkachenko: Why this word order? Wouldn't `isCampaignBotKeyword()` read better?

api/controllers/CampaignBotController.js
```diff
@@ -239,39 +234,51 @@ class CampaignBotController {
...
+    const configValue = process.env[configName];
+    if (!configValue) {
+      logger.error(`${this.loggerPrefix(req)} process.env.${configName} is undefined`);
 
       return false;
     }
 
-    const input = helpers.getFirstWord(req.incoming_message);
-    const isCommand = input && input.toUpperCase() === cmdClear.toUpperCase();
+    const isCommand = this.parseCommand(req) === configValue.toUpperCase();
```
sergii-tkachenko: It might be a bad idea to name the variable the same as the function. Kinda confusing.
In my opinion, simple `result` would suffice.

api/controllers/CampaignBotController.js
```diff
@@ -239,39 +234,51 @@ class CampaignBotController {
...
+    if (!type) {
+      return false;
+    }
 
-    if (req.user.role !== 'staff' && req.user.role !== 'admin') {
-      logger.warn(`${this.loggerPrefix(req)} unauthorized clear cache`);
+    const configName = `GAMBIT_CMD_${type.toUpperCase()}`;
+    const configValue = process.env[configName];
+    if (!configValue) {
+      logger.error(`${this.loggerPrefix(req)} process.env.${configName} is undefined`);
```
sergii-tkachenko: Why did you change it from warn to error?

api/controllers/CampaignBotController.js
```diff
@@ -239,39 +234,51 @@ class CampaignBotController {
...
 
       return false;
     }
 
-    const input = helpers.getFirstWord(req.incoming_message);
-    const isCommand = input && input.toUpperCase() === cmdClear.toUpperCase();
+    const isCommand = this.parseCommand(req) === configValue.toUpperCase();
+
+    if (type === 'clear_cache') {
+      const authorized = req.user.role && (req.user.role === 'staff' || req.user.role === 'admin');
```
sergii-tkachenko: There should be a function for that.

api/controllers/CampaignBotController.js
```diff
@@ -239,39 +234,51 @@ class CampaignBotController {
...
-   * @param {object} req - Express request
+   * Returns whether incoming request is the CampaignBot Mobile Commons keyword.
+   * @param {object} req
    * @return {bool}
    */
-  isCommandReportback(req) {
-    const firstWord = helpers.getFirstWord(req.incoming_message);
-    if (!firstWord) return false;
+  isKeywordCampaignBot(req) {
+    if (req.body.keyword) {
```
sergii-tkachenko:
```js
if (!req.body.keyword) {
  return false;
}

// Continue as normal
```

api/controllers/CampaignBotController.js
```diff
@@ -433,6 +447,9 @@ class CampaignBotController {
     msg = msg.replace(/{{rb_noun}}/gi, campaign.reportbackInfo.noun);
     msg = msg.replace(/{{rb_verb}}/gi, campaign.reportbackInfo.verb);
     msg = msg.replace(/{{rb_confirmed}}/gi, campaign.reportbackInfo.confirmationMessage);
+    msg = msg.replace(/{{cmd_reportback}}/gi, process.env.GAMBIT_CMD_REPORTBACK);
```
sergii-tkachenko: Is there going to be more than one match? If no, `/g` is an overhead.

config/router-chatbot.js
```diff
@@ -34,10 +37,19 @@ router.post('/', (req, res) => {
     return app.locals.controllers.donorsChooseBot.chatbot(req, res);
   }
 
-  const mobilecommonsOip = process.env.CAMPAIGNBOT_MOBILECOMMONS_OIP;
-  if (!mobilecommonsOip) {
-    logger.error('CAMPAIGNBOT_MOBILECOMMONS_OIP undefined');
+  let configured = true;
+  // Check for required config variables.
+  const settings = ['MOBILECOMMONS_OIP_CAMPAIGNBOT', 'MOBILECOMMONS_OIP_AGENTVIEW'];
+  settings.push('MOBILECOMMONS_KEYWORD_CAMPAIGNBOT');
```
sergii-tkachenko: Maybe just add it to `concat` below?

sergii-tkachenko: Or even just add all then to the array above.

config/router-chatbot.js
```diff
@@ -34,10 +37,19 @@ router.post('/', (req, res) => {
...
+  settings.concat(['GAMBIT_CMD_MEMBER_SUPPORT', 'GAMBIT_CMD_REPORTBACK']);
+  settings.forEach((configVar) => {
+    if (!process.env[configVar]) {
+      logger.error(`undefined process.env.${configVar}`);
+      configured = false;
```
sergii-tkachenko: Did you mean to end the loop when the first var is not found?

config/router-chatbot.js
```diff
@@ -132,12 +148,18 @@ router.post('/', (req, res) => {
...
+        oip = process.env.MOBILECOMMONS_OIP_AGENTVIEW;
+      }
+
+      mobilecommons.chatbot(req.body, oip, msg);
 
       return res.send({ message: msg });
     })
     .catch(err => {
       controller.error(req, res, err);
+
```
sergii-tkachenko: Extra line?

lib/helpers.js
```diff
@@ -133,6 +133,18 @@ function getFirstWord(message) {
 };
 module.exports.getFirstWord = getFirstWord;
 
+
+/**
+ * Returns the uppercase first word of given message.
+ */
+module.exports.getFirstWordUppercase = function (message) {
+  const isText = message || getFirstWord(message);
```
sergii-tkachenko: You shouldn't call `getFirstWord(message)` twice.

sergiitk (changes requested):

api/controllers/CampaignBotController.js
aaronschachter: I think this was forward thinking to an issue about to create about making use of the `body.keyword` property we get from a Mobile Commons request, where we'll have different keyword types. Signup, Continue. CampaignBot keyword here should really be continue.

I'm thinking this new function could work like the `isCommand(req, type)` does -- `isKeyword(req, type)` where type can be `'signup'` or `'continue'`.

config/router-chatbot.js
aaronschachter: I did write it that way at first -- but thought it could be useful to log each missing variable too.

config/router-chatbot.js
aaronschachter: Yeah this isn't as clean. I was debating on adding them all to 1 line adding using the `disable-eslint` -- would that be best @sergii-tkachenko ?

api/controllers/CampaignBotController.js
aaronschachter: Good call, there shouldn't be.

config/router-chatbot.js
aaronschachter: I was adding the extra line in attempt to make the `return` more readable (like above on L 158). Per http://eslint.org/docs/rules/newline-before-return -- there's no "hard and fast rule". I'm fine with removing, but maybe it could make sense for us to add one (a rule?)

sergii-tkachenko: Oh, good. I missed (forgot?) this rule. Thank you for pointing it out, will be doing the same from now on!

api/controllers/CampaignBotController.js
```diff
@@ -239,39 +234,46 @@ class CampaignBotController {
...
+      logger.warn(`${this.loggerPrefix(req)} process.env.${configName} is undefined`);
 
       return false;
     }
 
-    const input = helpers.getFirstWord(req.incoming_message);
-    const isCommand = input && input.toUpperCase() === cmdClear.toUpperCase();
+    const result = this.parseCommand(req) === configValue.toUpperCase();
+
+    if (type === 'clear_cache') {
```
sergii-tkachenko: Do you want to move this line condition above the `const result` line?
It looks like it doesn't rely on anything so we don't need to calculate result when the type is 'clear_cache', right?

api/controllers/CampaignBotController.js
```diff
@@ -442,7 +454,8 @@ class CampaignBotController {
       msg = msg.replace(/{{quantity}}/gi, quantity);
     }
 
-    if (req.query.campaign && req.signup && req.signup.draft_reportback_submission) {
+    const continuing = req.body.keyword && req.signup && req.signup.draft_reportback_submission;
```
sergii-tkachenko: Did you change it from `revisiting`? I thought it was more explicit.

sergiitk (changes requested):

api/controllers/CampaignBotController.js
aaronschachter: Ah ok i'll back to `revisiting` -- good point

api/controllers/CampaignBotController.js
aaronschachter: Great catch @sergii-tkachenko -- this was firing when it wasn't supposed to because we do need that `result` -- https://github.com/DoSomething/gambit/pull/654/commits/41cfc05d239832cd6f04ae50379454b394ae07ba

aaronschachter:
@sergii-tkachenko 👍  ?

sergiitk (approved):
👍 👍 👍

Issue #655, opened 2016-10-07
1 mData to rule them all: deprecate campaign query param, add Campaign.keyword property


aaronschachter:
## Background

To create a Mobile Commons Campaign Signup keyword ABC for a DS Campaign 123, current setup requires creating a new Mobile Commons mData that points to `chatbot?campaign=123` and has the ABC keyword defined on it:
<img width="793" alt="screen shot 2016-10-04 at 10 12 09 am" src="https://cloud.githubusercontent.com/assets/1236811/19199367/8fe94b1c-8c78-11e6-9b8b-f5e6f5b67473.png">

There are some major disadvantages to this approach:
* Configuration points of failure -- an mData has to be configured just right in order for it to work (make sure request method is set to POST, set the correct Gambit domain, set the correct`campaign` in the query string, make sure "Send all future replies" is checked, remember to include `x-gambit-api-key` by checking "Send custom headers")
* Can't easily change our API key -- need to bulk update all mData's (https://github.com/DoSomething/gambit/issues/576#issuecomment-241801209)
* More stuff / search results / option to choose from / general mess+clutter in our Mobile Commons admin

### Mobile Commons configuration 📱 

* [x] Define all Mobile Commons keywords on default `chatbot` mData, "Production CampaignBot"  (and Staging for staging keywords) -- the one that our `MOBILECOMMONS_OIP_CAMPAIGNBOT` triggers, posting to `/api/v1/chatbot` (no query params).

* [x] For each keyword you add, also add it to the corresponding Campaign model's `Campaign.keyword` property via Compose UI 

### Gambit 🤖 

* [x] Cache DS Campaigns in Gambit `config` db ([again](#653))

* [x] Add a `keyword` property to store Signup keyword for a given Campaign

* [x] Refactor Gambit to inspect the incoming `req.body.keyword` parameter instead of looking for a `req.query.campaign`. If we have a keyword, check what Campaign its for and load if the Campaign is active. 

### Side issues

* [x] Adding Campaigns back into config gives us a way to define CampaignBot properties on a Campaign itself to easily implement #652 

* [ ] Check if Campaign status is closed, add Gambit Jr support for a new `msg_campaign_closed`

* [ ] Create "Select Campaign" menu if we have trouble loading the campaign via keyword (or if your saved `current_campaign` is closed

Issue #652, opened 2016-10-05
Support to override CampaignBot content per Campaign


aaronschachter:
Currently using 1 CampaignBot for all Campaigns. Add a `campaignBot` property to a DS Campaign to load its content instead of using the default CampaignBot content.

PR #656, opened 2016-10-07
Deprecate req.query.campaign

aaronschachter:
#### What's this PR do?
- Hardcodes CampaignBot keyword / campaign map to deprecate query parameters
- We can use a single CampaignBot mData now that requests for all DS Campaigns go to a single `/v1/chatbot` URL.  For each incoming request, we determine which Campaign to load first by checking if the User texted in a CampaignBot keyword, if not we load the Campaign stored on the User's `current_campaign` property.

Locally : set expected `keyword` param and POST to `/v1/chatbot` to test conversation

Staging:
- [x] Deploy to staging
- [x] Deactivate all mData's with a `?campaign=` defined
- [x] Add each keyword from de-activated mData onto the CampaignBot mData's keywords
- [x] Test keywords, conversation

Next up is adding a Campaign model back in to avoid hardcoding and begin work on #652 
#### Relevant tickets
#655
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/config/router-chatbot.js
+++ b/config/router-chatbot.js
@@ -57,7 +57,27 @@ router.post('/', (req, res) => {
     return res.sendStatus(500);
   }
 
-  let campaignId = req.query.campaign;
+  let campaignId;
+  if (req.body.keyword) {
+    const keyword = req.body.keyword.toUpperCase();
+    switch (keyword) {
+      case 'BOOKBOT':
+        campaignId = 2299;
+        break;
+      case 'BUMBLEBOT':
+        campaignId = 2070;
+        break;
+      case 'LUCKYBOT':
+        campaignId = 2900;
+        break;
+      case 'SLEEVEBOT':
+        campaignId = 2938;
+        break;
+      default:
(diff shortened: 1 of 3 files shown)


config/router-chatbot.js
```diff
@@ -57,7 +57,21 @@ router.post('/', (req, res) => {
     return res.sendStatus(500);
   }
 
-  let campaignId = req.query.campaign;
+  let campaignId;
+  if (req.body.keyword) {
+    const keyword = req.body.keyword.toUpperCase();
+
+    if (keyword === 'BOOKBOT') {
```
sergii-tkachenko: Switch/case?

sergiitk (changes requested):

aaronschachter:
Adding screenshot of our CampaignBot mData to reference in https://github.com/DoSomething/gambit/wiki/Chatbot -- all keywords get added here now (refs #655)

screen shot 2016-10-07 at 1 54 11 pm

PR #632, opened 2016-09-28
CampaignBot router + refactor

aaronschachter:
#### What's this PR do?
To the end user, all this PR does sync an existing Signup from Northstar if you've already signed up for the DS Campaign XYZ, but it's your first time texting into the CampaignBot for XYZ.

It also refactors the monolithic `CampaignBotController.constructor` callback hell into smaller functions, and moves handling our Express response into `router.js` per @deadlybutter's suggestion in https://github.com/DoSomething/gambit/pull/631#discussion_r80523908

Plus:
* Refactor functions like `collectQuantity` and `collectPhoto` into a `collectReportbackProperty(req, property, ask)` function
* Adds a `getCurrentSignup` req to call Northstar JS Signups.index and check for existence of Signup. TODO -- Check for `Signup.campaignRun.current` right now it's hardcoded to just use first Signup returned
* Cleans up `/config/index.js` -- saves our User, Signup, and ReportbackSubmission models to `app.locals.db`, refactor Northstar and Phoenix clients to `app.locals.clients`
* Stores the current User and Signup in our Express request object instead of defining on `CampaignBotController` - refs https://github.com/DoSomething/gambit/pull/631/files#r80523498
* Creates `CampaignBotController` once in `server.js` to avoid unnecessarily creating/destroying -- refs https://github.com/DoSomething/gambit/pull/631/files#r80525317

#### How should this be reviewed?
Testing as an end user is one way... reviewing code is a whole lot trickier since this refactors the entire `CampaignBotController` class. I jumped off the deep end on this one, but it should make writing unit tests for these smaller functions a whole lot easier in the future.

End user testing needs a full test of all available routes (including legacy) -- since we're making some changes into main files like `server.js`, `router.js`, etc

Code review could start from the `campaignBot(req, res)` function in `router.js` and check out `CampaignBotController` classes accordingly.

#### Any background context you want to provide?

* Planning to move `/api/router.js` into `/config/router.js` - was going to give that a shot in this branch but it breaks the routes nested in `/api/legacy/`

* Removes check for `user.supports_mms` for now. This seems like something we may not need to store at all -- could be handled in our `msg_no_photo_sent` copy.

* It's so tempting to refactor the `DonorsChooseBotController` too.

#### Relevant tickets
#606

#### Checklist
- [x] Tested on staging.

cc @sergii-tkachenko

Issue #591, opened 2016-08-23
Archive / drop unused database collections


aaronschachter:
Killing legacy == no more `config` database, just `mdata_responder`. Here's what we won't use:

* [x] The `config` database

Tips (#542)
- [x] `mdata_responder.tips`

SMS Games (#525)
- [x] `mdata_responder.sg_competitivestory_games`
- [x] `mdata_responder.sg_gamecreateconfigs`
- [x] `mdata_responder.sg_gamemappings`
- [x] `mdata_responder.sg_users`

DonorsChoose -- 2016+ donations are stored in `donorschoose_donations`
- [x] `mdata_responder.donation_infos`

Legacy DS Campaigns (#706)

- [x] `reportbacks`
- [x] `logs_winston`
- [x] `logs_exceptions`

aaronschachter:
Once #706 is in -- we will no longer be using the `config` database at all.

It could be nice to move the `campaigns` collection to it so admin's aren't editing documents in our production DB with users, signups, etc - we'd just need to set up a staging `config` DB.

There's something to be said about having a single database though, so I'm inclined to just keep the `campaigns` in the `mdata_responder` for simplicity...

aaronschachter:
Dropped the remaining `reportbacks`, `logs_winston` and `logs_exceptions` from the `mdata_responder` database today upon deploying #706 /  [3.2.1](https://github.com/DoSomething/gambit/releases/tag/3.2.1) to production.

Can't figure out how to actually drop the `config` database from our Compose deployment, so I manually dropped all of the collections within it.

PR #657, opened 2016-10-10
Add Campaign model to store keyword and CampaignBot overrides

### LINKED ISSUE #652: Support to override CampaignBot content per Campaign

### LINKED ISSUE #655: 1 mData to rule them all: deprecate campaign query param, add Campaign.keyword property

aaronschachter:
#### What's this PR do?

Adds a Campaign model back into Gambit (removed in #632) to cache DS Campaigns, with additional properties:
- `keyword` -- Stores the DS Campaign's Mobile Commons keyword, to avoid hardcoding Campaign keywords (#656)
- Adds `msg_` CampaignBot properties to support overriding CampaignBot `msg`  content per Campaign (closes #652)

On app start, we query Phoenix to update each Campaign model defined in `CAMPAIGNBOT_CAMPAIGNS` with its latest Campaign data, and load the Campaign model into our  `app.locals.campaigns` hash map. Upon texting in a keyword, we check `app.locals.keywords` to check if it has an associated Campaign (and loads it if so).

Additional refactoring:
- Removes `MOBILECOMMONS_KEYWORD_CAMPAIGNBOT`  config in favor of rendering the selected Campaign's keyword in `CampaignBotController.renderResponseMessage(req, msgTxt)`
- Renames `MOBILECOMMONS_OIP_CAMPAIGNBOT` as `MOBILECOMMONS_OIP_CHATBOT` -- end game is to use a single mData for all SMS chatbot when possible.
- Refactors `app.controllers.campaignBots` hash map as a single `app.controllers.campaignBot` property. We load a single `CampaignBot` from Gambit Jr. to provide content for any DS Campaign -- overrides are handled via properties on the `Campaign` model (original d was to create a new `CampaignBot` instance in Gambit Jr)
- Removes `msg_ask_supports_mms` and `msg_no_supports_mms` for now, info for non-MMS members can be displayed within `msg_no_photo_sent`

I've added each staging Campaign's keyword into the `config` db's `campaigns` collection to test on our staging server.
- code: 👀 💻 
- integration: 📱 text each keyword defined in `config.campaigns` to signup/continue for its relevant Campaign
#### Relevant tickets
- Closes #655 
- Closes #652 
#### Checklist
- [x] Set `MOBILECOMMONS_OIP_CHATBOT` on staging, remove `MOBILECOMMONS_OIP_CAMPAIGNBOT`, `MOBILECOMMONS_KEYWORD_CAMPAIGNBOT`
- [x] Tested on staging.
- [ ] Set `MOBILECOMMONS_OIP_CHATBOT` on production


config/locals.js
```diff
@@ -33,59 +33,122 @@ function getModelMap(configName, model) {
...
+
+      return app.locals.controllers.donorsChooseBot;
+    });
+}
+
+/**
+ * Loads app.locals.db object of Mongoose models.
+ */
+function loadDb(conn) {
+  app.locals.db.campaigns = rootRequire('api/models/Campaign')(conn);
```
aaronschachter: Storing the Campaign model in our operations DB to allow defining keywords per environment (and [mData](https://github.com/DoSomething/gambit/wiki/Chatbot#mdata)).

The remaining collections in our `config` DB are only used by legacy endpoints -- meaning we can drop the `config` collection entirely in v3 (refs #591)

api/models/Campaign.js
```diff
@@ -0,0 +1,37 @@
...
+const mongoose = require('mongoose');
+
+const schema = new mongoose.Schema({
+
+  _id: { type: Number, index: true },
+  keyword: String,
+
+  // Properties cached from DS API.
+  title: String,
+  rb_confirmed: String,
```
sergii-tkachenko: Maybe `rb_confirmation_message`? I just thought why it's not boolean, then realized it's a message.

api/models/Campaign.js
```diff
@@ -0,0 +1,37 @@
...
+  rb_noun: String,
+  rb_verb: String,
+  status: String,
```
sergii-tkachenko: Would it make sense to add a [`enum`](http://mongoosejs.com/docs/api.html#schema_string_SchemaString-enum) validator? Not sure if that's the case here, but if it's restricted to certain values, it would make sense.

config/locals.js
```diff
@@ -33,59 +33,122 @@ function getModelMap(configName, model) {
...
+    rb_confirmed: campaign.reportbackInfo.confirmationMessage,
+    rb_noun: campaign.reportbackInfo.noun,
+    rb_verb: campaign.reportbackInfo.verb,
+  };
+
+  return app.locals.db.campaigns
+    .findOneAndUpdate({ _id: campaign.id }, data, { upsert: true, new: true })
+    .exec()
+    .then((campaignDoc) => {
+      app.locals.campaigns[campaign.id] = campaignDoc;
```
sergii-tkachenko: I'd add a basic validation for the `campaignDoc` var. Just to make sure that `_id` present.

sergiitk (approved):
Good! Minor comments if you're up to it.

api/models/Campaign.js
aaronschachter: @sergii-tkachenko This corresponds to the Campaign's `field_campaign_status` in Phoenix -- which has values `active` and `closed`.

sergii-tkachenko: Enum makes sense then.

sergiitk (approved):
👍

PR #665, opened 2016-10-15
Refactor CampaignBot parameter reassignment

### LINKED ISSUE #643: ESLint cleanup - remove no-param-reassign usage
Refs https://github.com/DoSomething/gambit/pull/641#discussion_r81379461

aaronschachter:
#### What's this PR do?
- Removes all `eslint-disable no-param-reassign`, creating a `scope` copy of `req` when we need to pass along objects like current User, Campaign, Signup in the `router-chatbot` and CampaignBot controller functions
- Refactors `req.campaign_id` calls to just use `req.campaign._id`
- Abstracts a `cacheSignup` function to be called upon getting from / posting to Signups API
- Fixes bug in #657 to save Signup.keyword if exists (only when member has signed up through SMS / Gambot)

code 👀 
staging 📲 
#### Relevant tickets

Fixes #643 
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -490,7 +490,7 @@ class CampaignBotController {
 
         const signup = req.signup;
         signup.reportback = rbid;
-        signup.total_quantity_submitted = req.signup.draft_reportback_submission.quantity;
+        signup.total_quantity_submitted = Number(req.signup.draft_reportback_submission.quantity);
         signup.updated_at = dateSubmitted;
         signup.draft_reportback_submission = undefined;
         return signup.save();
--- a/api/controllers/CampaignBotController.js
+++ b/api/controllers/CampaignBotController.js
@@ -168,8 +168,8 @@ class CampaignBotController {
 
     return app.locals.db.reportbackSubmissions
       .create({
-        campaign: scope.campaign._id,
-        user: scope.user._id,
+        campaign: req.campaign._id,
+        user: req.user._id,
       })
       .then(reportbackSubmission => {
         this.debug(scope, `created reportbackSubmission:${reportbackSubmission._id.toString()}`);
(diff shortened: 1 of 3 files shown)


config/router-chatbot.js
```diff
@@ -117,97 +117,95 @@ router.post('/', (req, res) => {
...
-        logger.info(`${controller.loggerPrefix(req)} cleared user.campaigns`);
+        scope.user.campaigns = {};
+        logger.info(`${controller.loggerPrefix(scope)} cleared user.campaigns`);
 
-        return controller.getCurrentSignup(req);
+        return controller.getCurrentSignup(scope);
       }
 
-      const signupId = user.campaigns[req.campaign_id];
+      // TODO: Sanity check. If we sitll haven't loaded scope.campaign here, push to the menu.
```
sergii-tkachenko: `sitll` = `still`

api/controllers/CampaignBotController.js
```diff
@@ -462,50 +481,59 @@ class CampaignBotController {
...
-        /* eslint-disable no-param-reassign */
-        req.signup.reportback = rbid;
-        req.signup.total_quantity_submitted = req.signup.draft_reportback_submission.quantity;
-        req.signup.updated_at = dateSubmitted;
-        req.signup.draft_reportback_submission = undefined;
-        /* eslint-enable no-param-reassign */
-        req.signup.save();
+        const signup = req.signup;
+        signup.reportback = rbid;
+        signup.total_quantity_submitted = req.signup.draft_reportback_submission.quantity;
```
sergii-tkachenko: Convert to int with `Number`?

api/controllers/CampaignBotController.js
```diff
@@ -462,50 +481,59 @@ class CampaignBotController {
...
+        signup.updated_at = dateSubmitted;
+        signup.draft_reportback_submission = undefined;
```
sergii-tkachenko: It was `null` in `cacheSignup()`.

api/controllers/CampaignBotController.js
```diff
@@ -462,50 +481,59 @@ class CampaignBotController {
...
-        req.signup.reportback = rbid;
-        req.signup.total_quantity_submitted = req.signup.draft_reportback_submission.quantity;
-        req.signup.updated_at = dateSubmitted;
-        req.signup.draft_reportback_submission = undefined;
-        /* eslint-enable no-param-reassign */
-        req.signup.save();
+        const signup = req.signup;
+        signup.reportback = rbid;
+        signup.total_quantity_submitted = req.signup.draft_reportback_submission.quantity;
+        signup.updated_at = dateSubmitted;
```
sergii-tkachenko: Format date?

sergiitk (changes requested):
It's actually good to go, just small notes.

api/controllers/CampaignBotController.js
aaronschachter: @sergii-tkachenko I'm specifically using `undefined` here to remove the `draft_reportback_submission` field from the Signup document, to indicate we don't have a draft Reportback Submission for this Signup.
screen shot 2016-10-17 at 10 35 39 am

Do you think it'd be cleaner to include the `draft_reportback_submission` property, with set to `null`? (which is what the behavior would be if we pass `null`)

sergiitk (approved):

Issue #562, opened 2016-08-06
Introduce standard logging


mshmsh5000:
_BUG_
-------------------
#### Current Behavior
This app logs messages to one MongoDB collection and exceptions to another. Because these aren't standard logging facilities, they don't

- Get rotated
- Funnel to ELK
- Trigger alerts
- Make themselves visible to anyone

#### Desired Behavior
- Find a standard log service that we can [funnel to ELK](https://www.google.com/search?q=heroku+logging+logstash).

cc @blisteringherb 

### TODO
Updated 10/18
* [x] Use one `logs` collection
* [x] Send to logstash

deadlybutter:
For reference https://devcenter.heroku.com/articles/log-drains

Just need to coordinate with Morgan on the credentials & testing it works

deadlybutter:
Related devops work: https://github.com/DoSomething/devops/issues/220

aaronschachter:
We're storing all logs in a `logs` table, and it sounds like we're good to go on Logstash. Potential last task would be to configure alerts if we don't have already -- Heroku sends emails about percentage of failed requests, and we've got New Relic configuration setup too.

@mshmsh5000 @blisteringherb Sound ok to close this one?

blisteringherb:
👍 We're also using PagerDuty now, so we might want to look into configuring that to escalate issues to the Gambit team.

PR #669, opened 2016-10-19
Deprecate logger.js

aaronschachter:
#### What's this PR do?
- Refactors `lib/logger.js` to create new `winston` class on `app.locals.logger` upon app launch
- Writes all winston logs to `logs` by default instead of 2 separate collections: `logs_winston` and `logs_exception`
- Test all current/legacy endpoints on staging

Looking to move all legacy code into a root `/legacy` directory -- we don't have much need for a separate `logger` wrapper class for `winston`
#### Relevant tickets

Starts on #562 
#### Checklist
- [x] Tested on staging.


sergii-tkachenko:
👍

Issue #673, opened 2016-10-20
Create Doing/Completed Mobile Commons Groups for a DS Campaign Run


aaronschachter:
Refs https://github.com/DoSomething/gambit/pull/672#issuecomment-255138491

When a new Campaign is cached, create [2 new Mobile Commons Groups](https://mobilecommons.zendesk.com/hc/en-us/articles/202052534-REST-API#createGroup) if they don't exist already with names
*  `campaign_[id]_run_[id]_doing` -- stores Users who haven't completed the Campaign yet
* `campaign_[id]_run_[id]_completed` -- stores Users who have completed the Campaign

We'll use these Mobile Commons Groups to segment our Mobile Commons broadcast messaging.

When a Gambit Signup document is created, add the User to the `signedup` Mobile Commons Group.
When a Gambit Signup is updated (which could happen from a POST #642) and the User has reported back, remove them from the `signedup` MC Group and add them to the `completed` Group.

aaronschachter:
We talked through how to implement today, and Quicksilver will handle adding/removing members to the relevant Doing/Completed groups. If Gambit handled this, Quicksilver would need to POST Reportback activity anyway (in order to move any Doing's into Completed if they reported back via web).

I was hoping we'd be able to use the Group machine name for adding/removing, but we need to pass the numeric Mobile Commons Group ID when [adding or removing](https://mobilecommons.zendesk.com/hc/en-us/articles/202052534-REST-API#AddGroupMember). 

One idea here is that when we add a new Campaign to Gambit's list of active campaigns (`CAMPAIGNBOT_CAMPAIGNS`) -- Gambit creates the groups if they don't already exist and could expose the ID's on a new GET Gambit `campaigns` endpoint exposing our campaign documents. 

Another idea could be for Quicksilver to [GET groups](https://mobilecommons.zendesk.com/hc/en-us/articles/202052534-REST-API#ListGroups) and parse the list looking for the correct group by its name (`campaign_id_run_id_doing`) results... but I think a GET Gambit `campaigns/:id` to find the correct Mobile Commons Groups sounds cleaner.

cc @sergii-tkachenko

aaronschachter:
This branch starts on creating the Mobile Commons Groups per DS Campaign upon app launch.  I could see a use case for also exposing the Campaign `keywords` and msg overrides defined in the Gambit document in a Gambit campaigns API as well.

[...]

PR #674, opened 2016-10-21
Creates mobile commons groups & saves to DB

deadlybutter:
#### What's this PR do?
- Adds some new packages
- Adds new fields to campaign model
- Creates mobile commons groups per campaign
- Saves ID's to model

LOL this is fun. So when testing in Mobile Commons, you cannot simple "delete" a group. You have to rename the group to some random string, then delete it. MoCo reserves the string in namespace even after you delete it.

Part 2 of this will be exposing this data

#### Relevant tickets
Fixes #

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/api/models/Campaign.js
+++ b/api/models/Campaign.js
@@ -71,7 +71,7 @@ schema.methods.createMobileCommonsGroups = function () {
   .then(() => mobilecommons.createGroup(`${prefix} status=completed`))
   .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
   .then(() => campaign.save())
-  .catch(err => logger.error(err););
+  .catch(err => logger.error(err));
 }
 
 module.exports = function (connection) {
--- a/api/models/Campaign.js
+++ b/api/models/Campaign.js
@@ -71,7 +71,7 @@ schema.methods.createMobileCommonsGroups = function () {
   .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
   .then(() => campaign.save())
   .catch(err => logger.error(err));
-}
+};
 
 module.exports = function (connection) {
   return connection.model('campaigns', schema);
--- a/api/models/Campaign.js
+++ b/api/models/Campaign.js
@@ -4,6 +4,8 @@
  * Models a DS Campaign.
  */
 const mongoose = require('mongoose');
+const mobilecommons = rootRequire('lib/mobilecommons');
+const parser = require('xml2json');
 
 const schema = new mongoose.Schema({
 
@@ -38,6 +40,37 @@ const schema = new mongoose.Schema({
   },
 
 });
+
+function setMobileCommonsGroup(campaign, status, group) {
+  const parsedGroup = JSON.parse(parser.toJson(group));
+
+  // If the group name is available...
+  if (parsedGroup.response.success === 'true') {
+
+    // Save newly created group id to this campaign.
+    const groupId = parsedGroup.response.group.id;
+    campaign.mobile_commons_groups[status] = groupId;
+  }
+}
+
+/**
+ * Create Doing/Completed Mobile Commons Groups to support Mobile Commons broadcasting.
+ * @see https://github.com/DoSomething/gambit/issues/673
+ */
+schema.methods.createMobileCommonsGroups = function() {
+  const campaign = this;
+  const prefix = `env=${process.env.NODE_ENV}
+                  campaign_id=${campaign._id}
+                  run_id=${campaign.current_run}`;
+
+  // Create mobile commons group with custom name based on this campaign
+  mobilecommons.createGroup(`${prefix} status=doing`)
+  .then(doingGroup => setMobileCommonsGroup(campaign, 'doing', doingGroup))
+  .then(() => mobilecommons.createGroup(`${prefix} status=completed`))
+  .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
+  .then(() => campaign.save())
+  .catch(err => logger.error(err));
+}
 
 module.exports = function (connection) {
   return connection.model('campaigns', schema);
--- a/api/models/Campaign.js
+++ b/api/models/Campaign.js
@@ -46,7 +46,6 @@ const schema = new mongoose.Schema({
 function setMobileCommonsGroup(campaign, status, group) {
   const scope = campaign;
   const parsedGroup = JSON.parse(parser.toJson(group));
-
   // If the group name is available...
   if (parsedGroup.response.success === 'true') {
 
@@ -72,7 +71,7 @@ schema.methods.createMobileCommonsGroups = function () {
   .then(() => mobilecommons.createGroup(`${prefix} status=completed`))
   .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
   .then(() => campaign.save())
-  .catch(err => logger.error(err));
+  .catch(err => logger.error(err););
 }
 
 module.exports = function (connection) {
--- a/api/models/Campaign.js
+++ b/api/models/Campaign.js
@@ -43,7 +43,8 @@ const schema = new mongoose.Schema({
 
 });
 
-function setMobileCommonsGroup(status, group) {
+function setMobileCommonsGroup(campaign, status, group) {
+  const scope = campaign;
   const parsedGroup = JSON.parse(parser.toJson(group));
 
   // If the group name is available...
@@ -51,7 +52,7 @@ function setMobileCommonsGroup(status, group) {
 
     // Save newly created group id to this campaign.
     const groupId = parsedGroup.response.group.id;
-    this.mobile_commons_groups[status] = groupId;
+    scope.mobile_commons_groups[status] = groupId;
   }
 }
 
@@ -67,9 +68,9 @@ schema.methods.createMobileCommonsGroups = function () {
 
   // Create mobile commons group with custom name based on this campaign
   mobilecommons.createGroup(`${prefix} status=doing`)
-  .then(doingGroup => setMobileCommonsGroup('doing', doingGroup).bind(campaign))
+  .then(doingGroup => setMobileCommonsGroup(campaign, 'doing', doingGroup))
   .then(() => mobilecommons.createGroup(`${prefix} status=completed`))
-  .then(completedGroup => setMobileCommonsGroup('completed', completedGroup).bind(campaign))
+  .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
   .then(() => campaign.save())
   .catch(err => logger.error(err));
 }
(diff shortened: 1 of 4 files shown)


config/locals.js
```diff
@@ -40,6 +43,44 @@ function cacheCampaign(phoenixCampaign) {
...
 
 /**
+ * Create Doing/Completed Mobile Commons Groups to support Mobile Commons broadcasting.
+ * @see https://github.com/DoSomething/gambit/issues/673
+ */
+function createMobileCommonsGroupsForCampaign(campaignModel) {
+  const campaignScope = campaignModel;
+  const prefix = `env=${process.env.NODE_ENV} campaign_id=${campaignModel._id}`;
+
+  // Migrate old campaign models.
```
aaronschachter: What is this here for?

deadlybutter:
Retested to make sure everything still works
- DB values
- Mobile Commons Groups

config/locals.js
```diff
@@ -40,6 +43,44 @@ function cacheCampaign(phoenixCampaign) {
 }
 
 /**
+ * Create Doing/Completed Mobile Commons Groups to support Mobile Commons broadcasting.
+ * @see https://github.com/DoSomething/gambit/issues/673
+ */
+function createMobileCommonsGroupsForCampaign(campaignModel) {
+  const campaignScope = campaignModel;
+  const prefix = `env=${process.env.NODE_ENV} campaign_id=${campaignModel._id}`;
```
aaronschachter: We'll want to include Campaign Run from https://github.com/DoSomething/phoenix-js/pull/16.

The hope is these Mobile Commons Groups are short-term fix for a more robust subscriber solution for Doing/Completed. But if we're still using this code in a year -- if we want to send a broadcast to all COMPLETED but re-open the campaign, we'll also need to clear the COMPLETED group -- if we add the Current Run ID into the prefix when we create the group, we avoid that scenario.

deadlybutter:
@aaronschachter was already planning on using that fancy new squash & merge button 😉

config/locals.js
deadlybutter: @aaronschachter 👍

config/locals.js
deadlybutter: Code was breaking on old DB models that didn't have `mobileCommonsGroups.completed`

deadlybutter:
@aaronschachter ready for a second look, heres the MoCo output! https://dosomething.mcommons.com/groups?utf8=%E2%9C%93&filter=env%3Dlocal

api/models/Campaign.js
```diff
@@ -12,6 +12,7 @@ const schema = new mongoose.Schema({
 
   // Properties cached from DS API.
   title: String,
+  run_id: Number,
```
aaronschachter: None of other Models use a `_id` naming convention -- e.g. `signup.user`, `signup.campaign`.  Let's rename this as `campaign_run`.

If there's a good argument to keep the `_id`, to make it  `campaign_run_id`-- we should change all other properties in other models to keep consistent.

api/models/Campaign.js
```diff
@@ -30,6 +31,12 @@ const schema = new mongoose.Schema({
   msg_menu_signedup: String,
   msg_no_photo_sent: String,
 
+  // Mobile Commons Specific Fields.
+  mobileCommonsGroups: {
```
aaronschachter: Same here -- all other models are using lowercase/underscore. If we're creating a `mobilecommons` object, I do consider whether `keywords` should be moved into it too.

config/locals.js
```diff
@@ -40,6 +44,46 @@ function cacheCampaign(phoenixCampaign) {
...
+  const prefix = `env=${process.env.NODE_ENV}
+                  campaign_id=${campaignModel._id}
+                  run_id=${campaignModel.run_id}`;
+
+  // Migrate old campaign models.
+  if (!campaignModel.mobileCommonsGroups.doing || !campaignModel.mobileCommonsGroups.completed) {
+    const groups = {
+      doing: '',
+      completed: '',
+    };
```
aaronschachter: Still don't get why we need this?

config/locals.js
```diff
@@ -40,6 +44,46 @@ function cacheCampaign(phoenixCampaign) {
 }
 
 /**
+ * Create Doing/Completed Mobile Commons Groups to support Mobile Commons broadcasting.
+ * @see https://github.com/DoSomething/gambit/issues/673
+ */
+function createMobileCommonsGroupsForCampaign(campaignModel) {
```
aaronschachter: I started on this function before figuring out how to use Mongoose instance method (implemented #675). This would be a perfect candidate to add as an [instance method](http://mongoosejs.com/docs/guide.html#methods) to our `Campaign` model .

package.json
```diff
@@ -42,11 +42,13 @@
     "path": "~0.4.9",
     "q": "^1.0.1",
     "request": "2.34.x",
+    "request-promise-native": "^1.0.3",
```
aaronschachter: I added `supergaent` initially to add Promises support, and because we use it Northstar-JS and Phoenix-JS. Not sure if added another library is anything to flag, but would be curious about @sergii-tkachenko 's opinion

aaronschachter:
Linking this PR to #673

config/locals.js
deadlybutter: Ah you're right we don't

package.json
deadlybutter: Request by default doesn't have promises, this package adds support for it.

api/models/Campaign.js
aaronschachter: I'm definitely in favor of keeping our naming conventions consistent across platforms, but agree nothing to change today. We do use `currentCampaignRun` in [Phoenix](https://github.com/DoSomething/phoenix-js/blob/7b4c45f7fe31bab7023d2408ce92f89b06d570fa/lib/phoenix-campaign.js#L16) -- my hesitation with `run` is it feels a little too generic, or questionable if it's actual DS Campaign Run. how about `current_run` ?

deadlybutter: kool 👍

api/models/Campaign.js
deadlybutter: I think a refactor that big should be a different PR

deadlybutter:
@aaronschachter wanna take another look at it? 👀

deadlybutter:
still testing something, disregard review request

deadlybutter:
I PASSED THE TEST. TAKE THAT LINT. THATS RIGHT. I WIN.

deadlybutter:
also @aaronschachter retested everything, its good for review

deadlybutter:
I see it in the app logs..

```
2016-10-24T17:05:58.001166+00:00 app[web.1]:                   campaign_id=2299
2016-10-24T17:05:58.007897+00:00 app[web.1]:                   run_id=6431 status=completed
2016-10-24T17:05:58.007896+00:00 app[web.1]:                   campaign_id=2273
2016-10-24T17:05:58.239370+00:00 app[web.1]: [34mdebug[39m: mobilecommons.createGroup env=thor
2016-10-24T17:05:58.239381+00:00 app[web.1]:                   campaign_id=2900
2016-10-24T17:05:58.250704+00:00 app[web.1]: [34mdebug[39m: mobilecommons.createGroup env=thor
2016-10-24T17:05:58.239382+00:00 app[web.1]:                   run_id=6477 status=completed
2016-10-24T17:05:58.250707+00:00 app[web.1]:                   campaign_id=2938
2016-10-24T17:05:58.250710+00:00 app[web.1]:                   run_id=6487 status=completed
```

Groups on thor created https://dosomething.mcommons.com/groups?utf8=%E2%9C%93&filter=env%3Dthor

Confirmed they're in the DB
screen shot 2016-10-24 at 1 20 09 pm


(Except for one, id 362, which according to the logs didn't load at startup https://gist.github.com/deadlybutter/3656874cf459246ec872dccf1b1134dc )

aaronschachter:
Something's off here. Running locally and the `mobilecommons_groups` aren't appearing -- also staging DB should contain `mobilecommons_groups`, not `mobile_commons_groups`.

api/models/Campaign.js
```diff
@@ -31,8 +35,51 @@ const schema = new mongoose.Schema({
...
 });
 
+/**
+ * For a given campaign, update its mobile commons group id,
+ * based on the mobile commons response.
+ * @param {Campaign} campaign    Campaign model to update.
+ * @param {string} status        Either completed or doing.
+ * @param {string} groupResponse XML string from mobile commons API.
+ */
+function setMobileCommonsGroup(campaign, status, groupResponse) {
```
aaronschachter: This function doesn't return anything

aaronschachter: https://github.com/DoSomething/gambit/pull/674/files#r84805483

aaronschachter:
My DB won't update itself upon app start since the Mobile Commons Groups exist already. It'd sure make things easier if the MC `create_group` API endpoint would return the existing Group ID back (or if we could query the `groups` endpoint by name)

deadlybutter:
@aaronschachter we talked about this -- Yeah I totally agree it's be sweet if they had a normal API, but until then this is the way we have to do it :/

api/models/Campaign.js
```diff
@@ -31,8 +35,51 @@ const schema = new mongoose.Schema({
...
+  const prefix = `env=${process.env.NODE_ENV}
+                  campaign_id=${campaign._id}
+                  run_id=${campaign.current_run}`;
+
+  // Create mobile commons group with custom name based on this campaign
+  mobilecommons.createGroup(`${prefix} status=doing`)
+  .then(doingGroup => setMobileCommonsGroup(campaign, 'doing', doingGroup))
+  .then(() => mobilecommons.createGroup(`${prefix} status=completed`))
+  .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
+  .then(() => campaign.save())
```
deadlybutter: Campaign is definitely updated, I think this is related to the problem we talked about below with you not being able to pull data.

deadlybutter:
Also remember you can't delete the group, you have to rename it then delete it! :facepalm:

api/models/Campaign.js
```diff
@@ -31,8 +35,51 @@ const schema = new mongoose.Schema({
...
+/**
+ * For a given campaign, update its mobile commons group id,
+ * based on the mobile commons response.
+ * @param {Campaign} campaign    Campaign model to update.
+ * @param {string} status        Either completed or doing.
+ * @param {string} groupResponse XML string from mobile commons API.
+ */
+function setMobileCommonsGroup(campaign, status, groupResponse) {
+  const scope = campaign;
+  const parsedGroup = JSON.parse(parser.toJson(groupResponse));
```
aaronschachter: @deadlybutter would it be cleaner to move parsing the XML response into the `mobileCommons.createGroup` function?

deadlybutter: I originally did that but for some reason the parsed response went back to XML when the caller function tried using the response. I got annoyed & moved it haha

aaronschachter: My 0.02 is it would be a lot cleaner to parse the JSON in `lib/mobilecommons.js` -- especially since we'll want to parse there too for a `getGroups` function last resort in https://github.com/DoSomething/gambit/issues/673#issuecomment-256251277

Issue #642, opened 2016-09-30
Send SMS Campaign Signup confirmation when member signs up via web and has Mobile Commons profile


aaronschachter:
This might be a way to support Signup confirmations from CampaignBot when the Signup is being created from a non-Gambit source (web, email, mobile app, etc).

* [x] Support posting a `signup_id` and `signup_source` to `chatbot`, to send a `chatbot` message to given Signup User for Signup Campaign
* [ ] Change Quicksilver to post our any new User registration or Signup creations to new Gambit endpoints directly instead of posting to Mobile Commons API, with `send_confirmation` set 
* [x] Add a new CampaignBot property named something like `menu_signedup_external` -- We want different signup confirmation copy for when a member signs up via SMS keyword vs. web. Rename current property as `menu_signedup_gambit`
* [x] Upon a POST Signups request, CampaignBot sends its `menu_signedup_external` message to the User
* [x] Add check for incoming POST requests -- only send a confirmation message if the Campaign is an active CampaignBot campaign defined in `CAMPAIGNBOT_CAMPAIGNS`

If we ever wanted to run 20 DS Campaigns at once over SMS, it brings up the many-signups-single-transactional confirmation QuickBit cc O'Doyle

cc @sergii-tkachenko @justkika  @classicfreddie  @jamjensen 

## How it works now

When a member signs up for a DS Campaign (id 362) on web and Campaign 362 has a unique Mobile Commons Opt-in Path (OIP) set, e.g. Staff Picks -- Quicksilver sends a signup confirmation SMS message to the member if we have their phone number.

Behind the scenes, Quicksilver is posting to the Mobile Commons API using that OIP to subscribe the member to a unique Mobile Commons Campaign we set up for current run of our DS Campaign 362.

The OIP we posted to sends a "You're signed up for Campaign Name! Let's get started" message to the member.  This copy is set by editing the OIP Conversation in the Mobile Commons admin UI.

Our DS Campaign's Mobile Commons Campaign contains several OIP's -- one for each type of message we send (Ask for quantity, Successful reportback, etc)

We also set up a unique Mobile Commons mData per DS Campaign name, which posts to our  `/reportback/:campaignName` legacy endpoint and used by all OIP's to send us incoming messages from the member.  Gambit legacy inspects the incoming message and responds back to the member by posting them into a different OIP (e.g. the OIP we have set for `ask_why` or `invalid_photo`)

## CampaignBot

[...]

PR #675, opened 2016-10-21
CampaignBot support to send confirmation messages for external signups

aaronschachter:
#### What's this PR do?

* Adds support to send Signup confirmation messages from CampaignBot for external (non Gambit) Signups:

    * Clients like Quicksilver should post to the `chatbot` endpoint and pass the `signup_id` and its `signup_source`

    * If the incoming `signup_source` is equal to the Gambit `DS_API_POST_SOURCE`, we don't send the confirmation message (we'll have already sent it to the User after they signed up through Gambit)

* Refactors `CampaignBotController` methods like `getUser` and `cacheSignup` as [Mongoose static and instance methods](http://mongoosejs.com/docs/guide.html#methods) in the relevant model classes

    * Adds Bluebird to promisify the new static/instance methods 

code 👀 
staging 📲 

#### Relevant tickets
#642 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -27,40 +27,10 @@ const userSchema = new mongoose.Schema({
 
 });
 
-module.exports = function (connection) {
-  return connection.model('users', userSchema);
-};
-
 /**
- * Statics.
+ * Parse given Northstar User for User model.
  */
-
-/**
- * Query DS API for given User type/id and store.
- */
-userSchema.statics.get = Promise.method((type, id) => {
-  logger.debug(`User.get type:${type} id:${id}`);
-
-  return app.locals.clients.northstar.Users
-    .get(type, id)
-    .then((northstarUser) => {
-      logger.debug('northstar.Users.get success');
-
-      return userSchema.statics.store(northstarUser);
-    })
-    .catch(() => {
-      logger.debug(`could not getUser type:${type} id:${id}`);
-
-      return null;
-    });
-});
-
-/**
- * Parse given Northstar User and return User model.
- */
-userSchema.statics.store = Promise.method((northstarUser) => {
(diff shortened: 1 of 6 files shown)


api/models/User.js
```diff
@@ -21,6 +27,69 @@ const schema = new mongoose.Schema({
...
+ * Instance methods.
+ */
+
+ /**
+ * Set user's current_campaign and save signup on campaigns hash map.
+ */
+UserSchema.methods.setCurrentCampaignWithSignup = Promise.method((signup) => {
+  logger.debug(`setCurrentCampaignWithSignup signup:${signup.id}`);
+
+  console.log(this._id);
```
aaronschachter: Undefined

config/router-chatbot.js
```diff
@@ -82,8 +84,63 @@ router.post('/', (req, res) => {
...
+        return app.locals.db.users.get('id', currentSignup.user);
+      })
+      .then((user) => {
+        if (!user) {
+          return res.status(500).status('Cannot find user for signup');
+        }
+
+        logger.debug(`user._id:${user._id}`);
+
+        return user.setCurrentCampaignWithSignup(currentSignup, () => {
```
aaronschachter: this works if the function is delcared htis way:
````
UserSchema.methods.setCurrentCampaignWithSignup = function(signup, callback) {
  console.log(this);
  console.log(signup);

  return true;
};
````

api/models/User.js
sergii-tkachenko: I think it's undefined because it's called within `Promise.method`. Just declare it as a normal method and return `new Promise`, see [this](https://github.com/DoSomething/phoenix-js/blob/master/lib/phoenix-client.js#L44) for an example.

PR #692, opened 2016-10-28
Always lookup User ID and current Signup ID, don't trust the cache

### LINKED ISSUE #679: Campaigns property doesn't get set for some users
Our CampaignBot mData reported a 500 error when @justkika (56ef3f40469c6444628b6688) triggered by texting into Don't Be A Sucker (id 46).

Looking into the db, there's no property set for user 56ef3f40469c6444628b6688's `campaigns` property for campaign 46, and the `current_campaign` is still set to 2070. Confirmed Campaign 46 is active. 

screen shot 2016-10-25 at 9 55 44 am

There's no Signup document for Campaign 46 in the `signups` collection:
screen shot 2016-10-25 at 9 58 18 am

### LINKED ISSUE #691: Lookup User ID for mobile number in every request (vs using stored Northstar ID MC Profile field)
The `chatbot` router currently checks [if we have a Northstar ID stored on the incoming Mobile Commons Profile](https://github.com/DoSomething/gambit/blob/develop/app/routes/chatbot.js#L72), to avoid fetching the User ID every time. We've already run into some bugs for internal staff members: The Northstar User ID on @hbghidey's Mobile Commons Profile (175753518) has this id stored for northstar ID: 55442907469c64ed7d8b480a. If you lookup that NS User, you'll see there's no mobile number, and a DS email. The Signup worked as expected, the Mobile Commons message was never delivered because we didn't have a mobile number for her user. Yet if you lookup a NS User for her mobile, its 5612a109469c64bc348b4f36 with a gmail address.

Here are ways a Northstar ID can exist on a Mobile Commons Profile:

* Quicksilver import scripts we ran for vCard to generate ID's and save to the field (to use within Liquid in a soon-to-be-legacy-MobileCommonsCampaign)

* Gambit updates it upon saving the `gambit_chatbot_response` Custom Field, [only when it doesn't exist](https://github.com/DoSomething/gambit/blob/develop/app/models/User.js#L134)

* Any Mobile Commons admin can edit this value via UI

I've seen 2 staffer instances that needed Northstar ID's cleared in their MC Profile to fix Gambot message failure, and suspicious it may be causing #679.

For CampaignBot v1 I think it's worth calling the extra GET users request to get Northstar ID for the incoming mobile number for every incoming mData request, as the incoming mobile number won't be so easy to change. It should also let us switch between testing Thor and Production Gambit environments without as much potential for ID mismatch.

cc @sergii-tkachenko @deadlybutter @DFurnes

aaronschachter:
#### What's this PR do?
- Refactors `app/routes/chatbot` to **GET** `/users` for the incoming `req.body.mobile` sent in each mData `chatbot` API request, instead of finding existing User document based on value saved to the `req.body.profile_northstar_id` field -- which was implemented in attempt to cut down on API requests. Any performance wins we gained by avoiding that GET request don't seem worth the inconsistent NS ID Profile value issues we're seeing in #691.
- Creates new user with only incoming `req.body.mobile` and `req.body.mobilecommons_id` values -- no longer post `first_name`, `email`, or `postal_code` if we have it. See "background context I want to provide".
- No longer saves Northstar ID to Mobile Commons Profile, as Gambit no longer needs it.
- Same for Signups: Always query for an ID, don't store anything to `User.campaigns` anymore. Only save the `user.current_campaign` upon successful request, so we know which campaign to load if user returns without a keyword included.
- Clear your Northstar ID profile field value, and text into Thor or Prod CampaignBot. Verify Northstar ID profile field is not updated (and that you're getting CampaignBot responses as expected)
- Should be able to freely text between Thor and Production campaigns without trouble -- which suggests it'd make sense to add either a `@thor: Hey it's me`prefix progamatically to all messages sent from Thor Gambit, or add/configure a separate new [Thor CampaignBot](http://dev-gambit-jr.pantheonsite.io/wp-json/wp/v2/campaignbots/41) to better distinguish which environment you're chatbotting with.

I experienced an odd bug today where saving my first name as "Go" in the DonorsChooseBot flow and then texting into production CampaignBot was causing my `user` to have a null phone number (yet my phone number and User ID were correct).  No idea why, but seems easier to just disregard any MC Profile info besides the critical info (`phone` being #1, `profile_id` seems reasonable to trust).

I initially included First Name, Email, and Postal Code profile fields in **POST** `users` thinking, "why not submit the info if we collect it for Science Sleuth?", but between HEROGATE (users providing false info to unlock donations) and this First Name mismatch bug I ran into -- seems best to [KISS](https://en.wikipedia.org/wiki/KISS_principle) for v1.
#### Relevant tickets
- Fixes #691 
- Fixes #679
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/models/Signup.js
+++ b/app/models/Signup.js
@@ -76,19 +76,59 @@ signupSchema.statics.lookupByID = function (signupID) {
   });
 };
 
-
-// TODO: DRY - when we move CampaignBotController getCurrentSignup to this class
-signupSchema.statics.storeNorthstarSignup = function (northstarSignup) {
+/**
+ * Gets current Signup for given User and Campaign from DS API, stores if found.
+ */
+signupSchema.statics.lookupCurrentForUserAndCampaign = function (user, campaign) {
   const model = this;
 
   return new Promise((resolve, reject) => {
-    logger.debug(`Signup.storeNorthstarSignup:${northstarSignup.id}`);
-    const data = parseNorthstarSignup(northstarSignup);
+    logger.debug(`Signup.lookupCurrentForUserAndCampaign:${user._id}, ${campaign._id}`);
 
-    return model.findOneAndUpdate({ _id: data._id }, data, { upsert: true, new: true })
-      .exec()
-      .then(signup => resolve(signup))
-      .catch(error => reject(error));
+    return app.locals.clients.northstar
+      .Signups.index({ user: user._id, campaigns: campaign._id })
+      .then(northstarSignups => {
+        if (!northstarSignups.length) {
+          return resolve(false);
+        }
+
+        // TODO: Loop through results to find result with northstarSignup.campaignRun.current.
+        const data = parseNorthstarSignup(northstarSignups[0]);
+
+        return model.findOneAndUpdate({ _id: data._id }, data, { upsert: true, new: true })
+          .populate('draft_reportback_submission')
+          .exec()
+          .then(signup => resolve(signup))
+          .catch(error => reject(error));
+      });
+  });
+};
+
+signupSchema.statics.post = function (user, campaign, keyword) {
+  const model = this;
+
+  return new Promise((resolve, reject) => {
+    logger.debug(`Signup.post (${user._id}, ${campaign._id}, ${keyword}`);
+
+    return app.locals.clients.phoenix.Campaigns
+      .signup(campaign._id, {
+        source: process.env.DS_API_POST_SOURCE,
+        uid: user.phoenix_id,
+      })
+      .then((signupID) => {
+        logger.debug(`Signup.post created signup:${signupID}`);
+        const data = {
+          campaign: campaign._id,
+          user: user._id,
+          keyword,
+        };
+
+        return model.findOneAndUpdate({ _id: signupID }, data, { upsert: true, new: true })
+          .exec()
+          .then(signup => resolve(signup))
+          .catch(error => reject(error));
+      })
+      .catch(err => reject(err));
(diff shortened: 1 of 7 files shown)


app/models/Signup.js
```diff
@@ -76,19 +76,59 @@ signupSchema.statics.lookupByID = function (signupID) {
...
+  const model = this;
+
+  return new Promise((resolve, reject) => {
+    logger.debug(`Signup.lookupCurrentForUserAndCampaign:${user._id}, ${campaign._id}`);
 
-// TODO: DRY - when we move CampaignBotController getCurrentSignup to this class
-signupSchema.statics.storeNorthstarSignup = function (northstarSignup) {
+    return app.locals.clients.northstar
+      .Signups.index({ user: user._id, campaigns: campaign._id })
+      .then(northstarSignups => {
```
sergii-tkachenko: Parentheses.

app/models/Signup.js
```diff
@@ -76,19 +76,59 @@ signupSchema.statics.lookupByID = function (signupID) {
...
+        if (!northstarSignups.length) {
```
sergii-tkachenko: > [15.](https://github.com/airbnb/javascript#comparison--shortcuts)3 Use shortcuts for booleans, but explicit comparisons for strings and numbers.

```js
if (northstarSignups.length < 1) {

}
```

sergiitk (changes requested):

aaronschachter:
Thanks @sergii-tkachenko ! Updated.

PR #697, opened 2016-10-28
Groups api

### LINKED ISSUE #673: Create Doing/Completed Mobile Commons Groups for a DS Campaign Run

deadlybutter:
#### What's this PR do?
- Adds gambit groups API lib
- Makes calls to Gambit Groups API. **Requires dot env file update!**
- Creates groups for campaigns if not present on the campaign

Good lord. Heres the TLDR, it works:
screen shot 2016-10-28 at 1 28 17 pm
screen shot 2016-10-28 at 1 28 24 pm

If you want to test this 100% through, you need to be really careful.
1. Delete the campaigns in your local DB.
2. Delete the group in the gambit groups prod DB.
3. Search mobile commons groups on there site for `campaign_id=some id`
4. Keep that tab open & open all 6 results into a new tab.
5. Rename each of those groups with some added random text. (We should probably start doing deleteme or some string that groups them together, sorry)
6. In that tab from step 3, select all of the groups & delete all.
7. Run Gambit.
HAVE FUN!

I need a cookie if you're gonna make me write this one out again.

#### Relevant tickets
Fixes #673 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/models/Campaign.js
+++ b/app/models/Campaign.js
@@ -96,17 +96,19 @@ campaignSchema.methods.createMobileCommonsGroups = function () {
   const campaignRunId = this.current_run;
 
   function parseApiRes(res) {
-    if (res._id) {
-      // If a group is in the response, save the data for this env.
-      const env = process.env.NODE_ENV;
-      const hasEnv = env in res.mobilecommons_groups;
-      return {
-        doing: hasEnv ? res.mobilecommons_groups[env].doing : '',
-        completed: hasEnv ? res.mobilecommons_groups[env].completed : '',
-      };
+    if (!res._id) {
+      return false;
     }
 
-    return false;
+    const env = process.env.NODE_ENV;
+    if (!res.mobilecommons_groups[env]) {
+      return false;
+    }
+
+    return {
+      doing: res.mobilecommons_groups[env].doing,
+      completed: res.mobilecommons_groups[env].completed,
+    }
   }
 
   gambitGroups.findGroup(campaignId, campaignRunId).then((res) => {
--- a/app/models/Campaign.js
+++ b/app/models/Campaign.js
@@ -103,7 +103,7 @@ campaignSchema.methods.createMobileCommonsGroups = function () {
       const env = process.env.NODE_ENV;
       return {
         doing: res.mobilecommons_groups[env].doing,
-        completed: res.mobilecommons_groups[env].completed
+        completed: res.mobilecommons_groups[env].completed,
       };
     }
 
@@ -122,7 +122,7 @@ campaignSchema.methods.createMobileCommonsGroups = function () {
     return gambitGroups.createGroup(campaignId, campaignRunId);
   })
   .then(res => {
-    if (res == undefined) {
+    if (res === undefined) {
       return;
     }
 
@@ -134,7 +134,7 @@ campaignSchema.methods.createMobileCommonsGroups = function () {
   })
   .then(() => {
     this.save();
-  })
+  });
 };
 
 module.exports = function (connection) {
--- a/app/models/Campaign.js
+++ b/app/models/Campaign.js
@@ -98,7 +98,8 @@ campaignSchema.methods.createMobileCommonsGroups = function () {
   function parseApiRes(res) {
     if (res._id) {
       // If a group is in the response, save the data for this env.
-      const hasEnv = process.env.NODE_ENV in res.mobilecommons_groups;
+      const env = process.env.NODE_ENV;
+      const hasEnv = env in res.mobilecommons_groups;
       return {
         doing: hasEnv ? res.mobilecommons_groups[env].doing : '',
         completed: hasEnv ? res.mobilecommons_groups[env].completed : '',
--- a/app/models/Campaign.js
+++ b/app/models/Campaign.js
@@ -4,9 +4,8 @@
  * Models a DS Campaign.
  */
 const mongoose = require('mongoose');
-const mobilecommons = rootRequire('lib/mobilecommons');
-const parser = require('xml2json');
 const logger = app.locals.logger;
+const gambitGroups = require('../../lib/groups');
 
 const campaignSchema = new mongoose.Schema({
 
@@ -59,23 +58,6 @@ function parsePhoenixCampaign(phoenixCampaign) {
   return data;
 }
 
-/**
- * For a given campaign, update its mobile commons group id,
- * based on the mobile commons response.
- * @param {Campaign} campaign    Campaign model to update.
- * @param {string} status        Either completed or doing.
- * @param {string} groupResponse XML string from mobile commons API.
- */
-function setMobileCommonsGroup(campaign, status, groupResponse) {
-  const scope = campaign;
-  const parsedGroup = JSON.parse(parser.toJson(groupResponse));
-  // If the group name is available...
-  if (parsedGroup.response.success === 'true') {
-    // Save newly created group id to this campaign.
-    const groupId = parsedGroup.response.group.id;
-    scope.mobilecommons_groups[status] = groupId;
-  }
-}
 
 /**
  * Get given Campaigns from DS API then store.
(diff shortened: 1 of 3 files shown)


app/models/Campaign.js
```diff
@@ -112,18 +94,47 @@ campaignSchema.statics.lookupByIDs = function (campaignIDs) {
...
-
-  // Create mobile commons group with custom name based on this campaign
-  mobilecommons.createGroup(`${prefix} status=doing`)
-  .then(doingGroup => setMobileCommonsGroup(campaign, 'doing', doingGroup))
-  .then(() => mobilecommons.createGroup(`${prefix} status=completed`))
-  .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
-  .then(() => campaign.save())
-  .catch(err => logger.error(err));
+  const campaignId = this._id;
+  const campaignRunId = this.current_run;
```
aaronschachter: Minor question -- is it more common in JS to use `whateverId` over `whateverID`, or just a preference thing? I've been using the capital `IDs` -- i'm happy to Replace All the ID's in a future PR but let's pick one as rule to keep consistent.

aaronschachter:
@deadlybutter --I think Gateway is looking for separate `mobilecommons_group_[status]` properties - https://github.com/DoSomething/gambit/issues/673#issuecomment-256251277

aaronschachter:
There's also a few tasks in that comment in #673 it'd make sense to split out into more issues (like a staging script)

app/models/Campaign.js
deadlybutter: I think we generally do `objectId`, heres an example i randomly found in Phoenix https://github.com/DoSomething/phoenix/blob/dev/lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js#L100

deadlybutter:
@aaronschachter hrm. I guess I can rename them but I would like to talk about grouping variables together to make the documents cleaner.

deadlybutter:
@aaronschachter updated & retested

app/models/Campaign.js
```diff
@@ -112,18 +92,47 @@ campaignSchema.statics.lookupByIDs = function (campaignIDs) {
...
+    const parsedRes = parseApiRes(res);
+    if (parsedRes) {
+      this.mobilecommons_group_doing = parsedRes.doing;
+      this.mobilecommons_group_completed = parsedRes.completed;
+      return undefined;
+    }
+
+    return gambitGroups.createGroup(campaignId, campaignRunId);
+  })
+  .then(res => {
```
sergii-tkachenko: Parentheses.

app/models/Campaign.js
```diff
@@ -112,18 +92,47 @@ campaignSchema.statics.lookupByIDs = function (campaignIDs) {
...
+    if (res === undefined) {
```
sergii-tkachenko: `if (!res)`?

app/models/Campaign.js
```diff
@@ -112,18 +92,47 @@ campaignSchema.statics.lookupByIDs = function (campaignIDs) {
...
-  .catch(err => logger.error(err));
+  const campaignId = this._id;
+  const campaignRunId = this.current_run;
+
+  function parseApiRes(res) {
+    if (res._id) {
+      // If a group is in the response, save the data for this env.
+      const env = process.env.NODE_ENV;
+      return {
+        doing: res.mobilecommons_groups[env].doing,
```
sergii-tkachenko: And what if it `env` isn't in `res.mobilecommons_groups`?

lib/groups.js
```diff
@@ -0,0 +1,65 @@
...
+const apiKey = process.env.GAMBIT_GROUPS_API_KEY;
+const apiHeaderName = 'x-gambit-group-api-key';
+
+/**
+ * Make a GET request to Gambit Groups API.
+ * Handles authentication & URL formation.
+ * @param  {string} path API Path. eg: /group
+ * @return {Promise}
+ */
+function get(path) {
```
sergii-tkachenko: Why is it outside of `module.exports`?

app/models/Campaign.js
deadlybutter: :thinking: that would be bad for a lot of reason! 

but I can do a safety check

app/models/Campaign.js
deadlybutter: sure

lib/groups.js
deadlybutter: I guess we can expose it, but I didn't see a need for it

deadlybutter:
retested again to make sure

lib/groups.js
sergii-tkachenko: You may want to call it from other module in this lib. Anyway, I see no reason to explicitly make them private.

app/models/Campaign.js
```diff
@@ -112,18 +92,48 @@ campaignSchema.statics.lookupByIDs = function (campaignIDs) {
...
-  .then(doingGroup => setMobileCommonsGroup(campaign, 'doing', doingGroup))
-  .then(() => mobilecommons.createGroup(`${prefix} status=completed`))
-  .then(completedGroup => setMobileCommonsGroup(campaign, 'completed', completedGroup))
-  .then(() => campaign.save())
-  .catch(err => logger.error(err));
+  const campaignId = this._id;
+  const campaignRunId = this.current_run;
+
+  function parseApiRes(res) {
+    if (res._id) {
```
sergii-tkachenko:
```js
if (!res._id) {
  return false;
}

const env = process.env.NODE_ENV;
if (!res.mobilecommons_groups[env]) {
  return false;
}

return {
  doing: res.mobilecommons_groups[env].doing,
  completed: res.mobilecommons_groups[env].completed,
}
```

deadlybutter: yeah that works 👍

lib/groups.js
```diff
@@ -0,0 +1,65 @@
...
+ * Handles authentication & URL formation.
+ * @param  {string} path API Path. eg: /group
+ * @param  {object} data JSON data to send.
+ * @return {Promise}
+ */
+function post(path, data) {
+  return superagent.post(`${uri}${path}`)
+    .send(data)
+    .set(apiHeaderName, apiKey)
+    .then(res => res.body)
```
sergii-tkachenko: Wait. Why do you call then/catch here? Just return the promise. You're thenning it above anyways.

deadlybutter: If I do it here I only have to access res.body once and handle the error once.

deadlybutter: So, both `findGroup` and `createGroup` would need there own response + error handlers. Which doesn't seem very dry to me

sergii-tkachenko: That's actually a bad idea to handle the error here. For example, you just want to handle different errors in different manners, like retry on timeout, exit on 500.

sergii-tkachenko: And regarding handling `res.body`: what if you wanted to check response headers in user code? Or HTTP?

deadlybutter: see my reply above ^ -- complications I have no reason to worry about right now

sergii-tkachenko: ok

lib/groups.js
```diff
@@ -0,0 +1,75 @@
...
+    return post('/group', {
+      campaign_id: campaignId,
+      campaign_run_id: campaignRunId
+    });
+  },
+
+  /**
+   * See docblock above.
+   */
+  get: get,
```
sergii-tkachenko: Why not just move func body here?

deadlybutter: was being lazy, sure i can move the entire thing

deadlybutter:
all tests passed again, any final thoughts on this @sergii-tkachenko ?

PR #700, opened 2016-10-28
Fixes Signups API, renames exception classes

### LINKED ISSUE #696: Signups endpoint fail
Removed `User.setCurrentCampaign` in #692 and forgot to test the [Signups endpoint.](https://github.com/DoSomething/gambit/blob/develop/app/routes/signups.js#L55)

Time to start thinking about some #671 soon.

aaronschachter:
#### What's this PR do?
- Fixes #696 and restores POST `signups` endpoint
- Renames `app/exceptions` as general classes

Test POST to `/v1/signups` with these params:
- id=2309254 source=`'mobileapp_ios'` --  Signup 2309254 is for my user for Campaign 46, currently running on staging CampaignBot. Shoudl return 200 with the rendered CampaignBot `msg_menu_signedup_external` message. Please use with care -- this will send a text to my number 📲  Test sending to your own phone by passing a Signup for your user and for a CampaignBot campaign, with a non `'sms-mobilecommons'` source.
- id=2309254 source=`'sms-mobilecommons'` - Should return a 208 for already sent, Signups with source `'sms-mobilecommons'` should not trigger sending a SMS confirmation
- id=1978786 source=`'mobileapp_ios'` - Should return 422, it's a Signup for a Campaign not running on CampaignBot 
- id=`'abc'` source=`'mobileapp_ios'` - Should return 404
#### Relevant tickets

Fixes #696
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/signups.js
+++ b/app/routes/signups.js
@@ -2,6 +2,7 @@
 
 const express = require('express');
 const router = express.Router(); // eslint-disable-line new-cap
+const NotFoundError = require('../exceptions/NotFoundError');
 const UnprocessibleEntityError = require('../exceptions/UnprocessibleEntityError');
 const logger = app.locals.logger;
 
@@ -82,6 +83,7 @@ router.post('/', (req, res) => {
 
       return sendResponse(res, 200, chatbotMessage);
     })
+    .catch(NotFoundError, err => sendResponse(res, 404, err.message))
     .catch(UnprocessibleEntityError, err => sendResponse(res, 422, err.message))
     .catch(err => sendResponse(res, 500, err.message));
 });
--- a/app/routes/signups.js
+++ b/app/routes/signups.js
@@ -2,7 +2,8 @@
 
 const express = require('express');
 const router = express.Router(); // eslint-disable-line new-cap
-const CampaignNotFoundError = require('../exceptions/CampaignNotFoundError');
+const NotFoundError = require('../exceptions/NotFoundError');
+const UnprocessibleEntityError = require('../exceptions/UnprocessibleEntityError');
 const logger = app.locals.logger;
 
 function successResponse(msg) {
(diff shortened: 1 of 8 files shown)


aaronschachter:
This PR should be good to go. Tested all cases listed in issue description and looks 👌

app/routes/signups.js
```diff
@@ -2,75 +2,90 @@
...
 
-  logger.debug(`POST /v1/signups { id:${signupID}, source:${source} }`);
+  logger.info(`POST signup:${signupId} source:${source}`);
 
   if (source === process.env.DS_API_POST_SOURCE) {
-    const msg = `Already sent confirmation for signup ${signupID} with source ${source}.`;
+    const msg = `CampaignBot only sends confirmation when source not equal to ${source}.`;
 
-    return res.send(successResponse(msg));
+    return sendResponse(res, 208, msg);
```
aaronschachter: @sergii-tkachenko Does Quicksilver blindly post Signups for CampaignBot Campaigns regardless of source? I'm assuming yes -- so returning a 208 vs an error code. If QS was only posting where source != `sms-mobilecommons`, then we would want to return an error code, indicating that check isn't working. It'd be cleaner to do the check to avoid unnecessary API lookup requests, but nothing we need for launch either.

sergii-tkachenko: Yes, but we can change that.

sergii-tkachenko: See [`MBC_RegistrationMobile_Service_MobileCommons::shouldBeProcessedOnGambit()`](https://github.com/DoSomething/mbc-registration-mobile/pull/47/files#diff-71e47cffa77bd2a3b2ca72c25a2f16b2R228).

aaronschachter: I think it'd make sense to make the change and throw an error code @sergii-tkachenko -- what do you think?

PR #709, opened 2016-11-02
Lookup current Signup in Signups.lookupCurrent for real

### LINKED ISSUE #639: CampaignBot: Add current Campaign Run validation in get/loadCurrentSignup
CampaignBot is currently returning the 1st found Signup for a given Campaign / User for sake of shipping a testable prototype, and [not checking the Signup is for the Campaign's current Run.
](https://github.com/DoSomething/gambit/blob/59b2f49750807de65c219cc4de30fa471cb4b298/api/controllers/CampaignBotController.js#L174)
Inspect the Signup's Campaign Run values to determine if it is no longer current and a new one should be created.

### LINKED ISSUE #710: Add Thor message prefix for staging CampaignBot conversations
For sanity's sake and internal use, let's prefix all messages coming from the staging CampaignBot with something like:
> thor: Hey thanks for signing up!

to better indicate when we're interacting with staging vs production CampaignBot mData's. Also could help us out big time to troubleshoot edge-cases where a member accidentally texts in a staging keyword -- if this happened, their signup/reportback (or even user) would be created on Thor and not production.

aaronschachter:
#### What's this PR do?
* Renames `Signup.lookupCurrentForUserAndCampaign(user, campaign)` function to just `Signup.lookupCurrent(user, campaign)`

* Adds check in `lookupCurrent` to return the first Signup found where `currentRun.current`, instead of just blindly returning the first Signup found. This was always a TODO for later -- and [that day has come, as Don't Be A Sucker is running on CampaignBot in its 2nd run](https://dosomething.slack.com/archives/gambit/p1478037777001803)

* Removes check for deprecated Clear Cache command. We no longer cache what the User's current signup is (removed in #692)

* Adds a new `GAMBIT_CHATBOT_RESPONSE_PREFIX` to as an environment indicator. Will set it to `@thor:` for staging to close #710

##### Current Signup

* Signup for a CampaignBot Campaign via staging SMS keyword.
* In Phoenix Thor, close the Campaign you signed up for and re-open it on a new run.
* Signup again for the Campaign via staging keyword -- verify a new Signup ID is created.

screen shot 2016-11-01 at 8 46 59 pm

##### Environment indicator / Chatbot Response Prefix

I've deployed this to staging -- should get messages back with a prefix of `@thor:` upon texting a  staging keyword.

slack for ios upload

#### Relevant tickets
* Fixes #639 
* Fixes #710 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -350,6 +350,11 @@ class CampaignBotController {
       msg = `${continueMsg} ${campaign.title}...\n\n${msg}`;
     }
 
+    const senderPrefix = process.env.GAMBIT_CHATBOT_RESPONSE_PREFIX;
+    if (senderPrefix) {
+      msg = `${senderPrefix} ${msg}`;
+    }
+
     return msg;
   }
 
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -187,33 +187,7 @@ class CampaignBotController {
 
     const configName = `GAMBIT_CMD_${type.toUpperCase()}`;
     const configValue = process.env[configName];
-    if (!configValue) {
-      logger.warn(`${this.loggerPrefix(req)} process.env.${configName} is undefined`);
-
-      return false;
-    }
-
     const result = this.parseCommand(req) === configValue.toUpperCase();
-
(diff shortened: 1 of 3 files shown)


sergiitk (approved):
👍

PR #716, opened 2016-11-03
Removes old HTTP max sockets code

deadlybutter:
Going wayyy back in the days of Node, max sockets was capped at 5. Developers had to manually increase this. However by v0.12, this was fixed to always be infinity. https://nodejs.org/en/blog/release/v0.12.0/

By capping this at 100, we're severely hurting ourselves, and it seems to correlate with the requests per minute chart in Heroku, where we keep taking large batches of ~100 requests and then do almost nothing

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: node --prof server.js
+web: npm start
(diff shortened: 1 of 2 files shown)


aaronschachter:
Fixin' up that ol 2-year-old-code. Looks ok to me (as long as the commits are squashed) -- @sergii-tkachenko ?

PR #719, opened 2016-11-04
Some small cleanup work around groups

### LINKED ISSUE #708: Groups cleanup
* [x] [Add `lib/groups.js` for linting](https://github.com/DoSomething/gambit/blob/02c84f8b346c838c157ea3a31552d2db4cc3ab17/package.json#L13)
* [x] Log err.message if DS Messaging API key doesn't exist instead of empty error -- see attachments
* [x] Remove `mobilecommons.createGroup`function if we're no longer using it (we're not, right? the DS Messaging Groups API handles?)
* [ ] DS Messaging Groups API: If the create group call fails, query the GET groups by name to find its ID (new MC API feature per support ticket #124466)
* [ ] Lower priority -- but DRY cleanup? My $0.02 is if code can be cleaner, we should clean it: https://github.com/DoSomething/gambit/pull/697/files/1f88a857619cecf03104bc078b93b9c71fbc1241#r85604276

cc @sergii-tkachenko 

## Attachments
screen shot 2016-10-29 at 8 07 23 am

deadlybutter:
#### What's this PR do?
- Adds groups.js to lint
- Logs error if missing API key or URI
- Simplifies some error handling code

- Does the groups API still return a response?

- nope

#### Relevant tickets
Fixes #708

### DIFF
--- a/lib/groups.js
+++ b/lib/groups.js
@@ -11,6 +11,10 @@ const uri = process.env.DS_MESSAGING_GROUPS_API_URI;
 const apiKey = process.env.DS_MESSAGING_GROUPS_API_KEY;
 const apiHeaderName = 'x-messaging-group-api-key';
 
+if (!uri || !apiKey) {
+  logger.error("Missing DS Messaging config vars");
+}
+
 module.exports = {
   /**
    * Make a GET request to Gambit Groups API.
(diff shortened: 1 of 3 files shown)


deadlybutter:
- added config var to stage + prod
- tested stage keywords

PR #584, opened 2016-08-18
DonorsChooseDonationController cleanup

aaronschachter:
#### What's this PR do?
* Fixes a bug setting the Donation Count profile field upon first donation
* Adds  `DONORSCHOOSE_BOT_ID` and `DONORSCHOOSE_MOCO_CAMPAIGN_ID` constants set from environment variables, stores in the `donorschoose_donations` collection
* Adds comments for functions, general code cleanup  (refs #573)

#### How should this be reviewed?
* Remove the value from your Donation Count profile field (we're using `ss2016_donation_count` for this year's run) and then test locally with this branch.
* Confirm MoCo Campaign ID and DonorsChoose Bot ID are stored upon successful donation

#### Relevant tickets
Fixes #581, #582 

#### Checklist
- [x] Tested on staging.

PR #720, opened 2016-11-04
Begins DonorsChooseBot refactor

aaronschachter:
#### What's this PR do?
Begins a `/v1/donorschoosebot` endpoint to eventually deprecate use of `/v1/chatbot?bot_type=donorschoosebot`. This endpoint will post a donation if all the expected Mobile Commons Profile fields are passed in the `req.body`, but doesn't save anything to the user's  Mobile Commons Profile. Opening this now to help keep the pull request reviewable.

* Adds new `app/routes/donorschoosebot` to handle logic of inspecting Mobile Commons Profile, sent message from User, and determining how to respond.
    * Uses promises to avoid callback hell (similar to CampaignBot makeover in #632)

* Adds `lib/donorschoose.js` for lint -- adds helper `superagent` functions for DonorsChoose API requests

* Adds `DONORSCHOOSE_API_BASEURI` and `DONORSCHOOSE_API_SECURE_BASEURI` to avoid hardcoding URL's for use with new route

Post to `/v1/donorschoosebot` locally with expected params defined in [chatbot docs](https://github.com/DoSomething/gambit/blob/develop/documentation/endpoints/chatbot.md), if verify ask/invalid messages are sent, and a donation is created if all fields are passed.

The only time we post to DonorsChooseBot is when a user has completed a SMS Game, so it makes sense to move it out of our `chatbot` endpoint entirely and into its own `donorschoosebot` endpoint.

Next steps:
* [ ] Save data to Mobile Commons Profile and deliver messages
* [ ] Update documentation
* [ ] Once live, remove the `DonorsChooseBotController`. Our `package.json` can simply lint our entire `app` directory instead of specifying each subdirectory

#### Relevant tickets
#584 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/donorschoosebot.js
+++ b/app/routes/donorschoosebot.js
@@ -5,6 +5,8 @@
  */
 const express = require('express');
 const router = express.Router(); // eslint-disable-line new-cap
+const superagent = require('superagent');
+const donorschoose = require('../../lib/donorschoose');
 const helpers = require('../../lib/helpers');
 const logger = app.locals.logger;
 
@@ -78,7 +80,24 @@ router.post('/', (req, res) => {
   // We've made it this far, time to donate.
   // TODO: When time to post to MC Profile, we'll need to save email if !req.body.profile_email.
 
-  return res.send('hi');
+  const donationAmount = (process.env.DONORSCHOOSE_DONATION_AMOUNT || 10);
+  const olderThan = process.env.DONORSCHOOSE_PROPOSALS_OLDERTHAN;
+  const zip = req.body.profile_postal_code;
+  // Find a project for zip code that has at least DONATION_AMOUNT left to completion.
+  const proposalsUri = donorschoose.getProposalsQueryUrl(zip, donationAmount, olderThan);
+  logger.debug(proposalsUri);
+
+  return superagent.get(proposalsUri).set('Accept', 'application/json')
+    .then((response) => {
+      const body = JSON.parse(response.text);
+      if (body.proposals.length < 1) {
+        logger.error(`'No proposals found for zip ${zip}`);
+        // TODO: Inform User, either by throwing an error or adding a no projects found message.
+      }
+      const project = body.proposals[0];
+      return res.status(200).send(project);
+    })
+    .catch(err => res.status(err.status).send(err.message));
 });
 
 module.exports = router;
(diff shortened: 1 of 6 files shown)


app/routes/donorschoosebot.js
```diff
@@ -0,0 +1,182 @@
...
+  let incomingMessage = req.body.args;
+  if (!incomingMessage) {
+    return helpers.sendResponse(res, 422, 'Missing required args.');
+  }
+
+  incomingMessage = helpers.getFirstWord(incomingMessage);
+  logger.debug(`dcbot incomingMessage:${incomingMessage}`);
+
+  const profileFieldName = `profile_${COUNT_FIELD}`;
+  const numDonations = req.body[profileFieldName];
```
sergii-tkachenko: If `req.body[profileFieldName] == undefined`?

app/routes/donorschoosebot.js
```diff
@@ -0,0 +1,182 @@
...
+  if (numDonations >= MAX_DONATIONS_ALLOWED) {
+    return sendResponse(res, 200, 'max_donations_reached');
+  }
+
+  // If start param passed, initiate conversation, ignoring incomingMessage.
+  const prompt = req.query.start;
+
+  if (!req.body.profile_postal_code) {
+    if (prompt) {
+      return sendResponse(res, 200, 'ask_zip');
```
sergii-tkachenko: Why 200? It's an error, right? Maybe 422 or 400?

sergiitk (changes requested):

app/routes/donorschoosebot.js
aaronschachter: It's not meant to be an error -- Mobile Commons posts an entire Profile in `req.body` -- this  check is to see if there's a value stored in that field. If not, we send a message to User asking for zip code. Send a 200 code back to Mobile Commons mData saying Yup I know how to respond to this, I sent a `profile_update` request.

sergii-tkachenko: Maybe then just explain it in the comment.

app/routes/donorschoosebot.js
sergii-tkachenko: I thought more like
```js
const numDonations = req.body[profileFieldName] ? Number(req.body[profileFieldName]) : -1;
```

sergiitk (approved):
👍

Issue #706, opened 2016-11-01
Remove legacy reportback endpoint


aaronschachter:
Starting Nov 9, this endpoint won't be used anymore. Once no longer used in prod:

* [x] Remove all files in `legacy` directory
* [x] Remove `CONFIG_DB_URI` config variable -- we no longer use the `config` database
* [x] Remove legacy phoenix `DS_CONTENT_API_` config variables 
* [x] Remove deprecated code in [`config/locals`](https://github.com/DoSomething/gambit/blob/479b1b09ced0c3cd78ee1ae533a256b5ff5776db/config/locals.js#L45)
* [x] Remove unused `mobilecommons` functions

aaronschachter:
Removed deprecated config variables from production. RIP, multiple OIP's.

PR #734, opened 2016-11-10
Removes legacy /reportback endpoint

aaronschachter:
#### What's this PR do?
* Removes the `/reportback/:campaignName` endpoint, now deprecated in favor of using a single Chatbot Mobile Commons Campaign and mData.
* Removes unused functions from `lib/helpers`
* Removes unused packages from `package.json`

* Verify entire CampaignBot and DonorsChooseBot conversations work as expected
* Verify `/reportback/:campaignName` endpoints return a 404. This will likely cause a few more Mobile Commons emails for 404's until we de-activate all the mData's that we no longer use.

Last item left to do here is add the entire `lib` directory to `scripts.lint` in our `package.json` and lint the legacy `lib/helpers.js` 

#### Relevant tickets
Very big dent in #706

#### Checklist
- [x] Tested on staging.


lib/helpers.js
```diff
@@ -122,7 +57,7 @@ module.exports.isValidPhone = function(phone) {
  *
  * @return First word of the message.
  */
-function getFirstWord(message) {
+module.exports.getFirstWord = function (message) {
```
aaronschachter: #738

PR #733, opened 2016-11-09
BotRequest model

aaronschachter:
#### What's this PR do?
Takes our Gambit Stathat events for bot messages in #726 a step further by storing them in a new `bot_requests` collection in the Gambit DB.

*  Log incoming POST requests to `/chatbot`, `/donorschoosebot`, and `/signups` along with the Bot response back:
    * `bot` -- The Bot that responded (`campaignbot`, `donorschoosebot`, future Onboarding, SMS Games, Account Management Bots that we need CMS's like Gambit Jr to maintain copy in)
   * `bot_msg_type` -- The type of Bot message we're sending back as a response, e.g. `'ask_zip'`, `'menu_completed'`)
    * `bot_msg` -- The rendered response message to deliver back to the User

* Adds a `client` property to our incoming Express request to determine which body properties to store.

    * `mobilecommons` -- incoming Mobile Commons mData request. If set, we store things like the `broadcast_id`, which is critical for Signup Broadcast Support #651 to gather analytics for how many Users signup from a given Campaign Signup Broadcast (filter by `body.broadcast_id` and `bot_response_type === 'msg_menu_signedup_gambit`)

    * `external-signup`  -- incoming Quicksilver request to trigger the CampaignBot External Signup Confirmation message (`msg_menu_signedup_external`)

#### How should this be reviewed?
1. I could use some second opinions on all names (both of `bot_requests` and all its fields), and exactly what messages we want to store. Am I missing anything?  Do the names make sense?

2. Currently logging all triggered External Signup confirmation messages -- but we sure send thousands and thousands of them. Still, seems like it would be odd not to. It would also help us catch if we happened to send multiple external confirmation messages if there were bugs.

3. I'll deploy to staging, and post around the various endpoints to start populating the table. Please take a look at live staging  `bot_requests` data and confirm 🆗 . Need access to the staging DB? There's [an app for that](https://github.com/DoSomething/ServerConfig/wiki/4.0-Gambit#staging).

#### Relevant tickets
Fixes #698 

#### Checklist
- [x] Tested on staging.

cc @deadlybutter @mshmsh5000 @sheyd

PR #739, opened 2016-11-11
Fixes isYesResponse bug from #733

### LINKED ISSUE #738: Signup Broadcast 500 error: getFirstWord is not defined
Responding to a Signup Broadcast results in an error introduced by #733.

* [x] Fix bug
* [ ] Document list of all endpoints / scenarios to fully test all CampaignBot scenarios / responses to avoid missing errors like #736 and this one. Refs #671

aaronschachter:
#### What's this PR do?
* Fixes bug introduced by #733 that throws 500 error upon responding to a Signup Broadcast
* Adds a just a little bit more logging to log when we call `isCommand` and if current request is a response to a Signup Broadcast
* Handles scenario where Signup Broadcast is sent to a User who's already signed up and has a draft Reportback Submission started. Answering Yes would result in the `invalid_zip` response -- instead we send back the next question to ask with the "Picking up where you left off" copy.

* Post to staging ~~`chatbot?broadcast=true`~~ `chatbot` endpoint and include a `broadcast_id` in the request body.
* Verify passing  yes and no responses return either the relevant Signed Up message, or the Signup Broadcast Declined message.

Think it makes sense to do a minor refactor here -- instead of checking for a `req.query.broadcast` parameter, look for a `req.body.broadcast_id` that gets sent when User is responding to a Mobile Commons broadcast. That would mean we can avoid use of a separate mData for broadcasts entirely, and just use the single Chatbot mData (POST `/v1/chatbot`) for both Chat OIP responses and Broadcast responses.

#### Relevant tickets
Fixes #738

#### Checklist
- [ ] Documentation added for new features/changed endpoints -- list all endpoints to test and scenarios
- [x] Tested on staging.

### DIFF
--- a/app/models/CampaignBot.js
+++ b/app/models/CampaignBot.js
@@ -71,8 +71,7 @@ campaignBotSchema.methods.renderMessage = function (req, msgType, prefix) {
   let msg = this[botProperty];
   const campaign = req.campaign;
   if (!campaign) {
-    stathat('campaignbot: undefined campaign');
-    logger.error('renderMessage req.campaign undefined');
+    logger.debug('renderMessage req.campaign undefined');
 
     return msg;
   }
--- a/app/models/CampaignBot.js
+++ b/app/models/CampaignBot.js
@@ -63,7 +63,7 @@ campaignBotSchema.statics.lookupByID = function (id) {
  * @return {string} - CampaignBot message with Liquid tags replaced with req properties
  */
 campaignBotSchema.methods.renderMessage = function (req, msgType, prefix) {
-  const logMsg = `campaignbot: ${msgType}`;
+  const logMsg = `campaignbot:${msgType}`;
   logger.info(logMsg);
   stathat(logMsg);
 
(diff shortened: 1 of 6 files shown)


aaronschachter:
Makes the change to check for the request `body.broadcast_id` instead of `query.broadcast_id` -- which means we no longer need 2 mData's (one for Chatbot, one for Signup Broadcasts with the query string) -- we can back to using just the 1 Chat mData.

app/models/CampaignBot.js
```diff
@@ -63,16 +63,15 @@ campaignBotSchema.statics.lookupByID = function (id) {
...
-  const logMsg = `campaignbot: ${msgType}`;
+  const logMsg = `campaignbot:${msgType}`;
   logger.info(logMsg);
   stathat(logMsg);
 
   const botProperty = `msg_${msgType}`;
   let msg = this[botProperty];
   const campaign = req.campaign;
   if (!campaign) {
-    stathat('campaignbot: undefined campaign');
```
aaronschachter: See https://github.com/DoSomething/gambit/pull/737/files#r87543988

sergiitk (approved):
👍

Issue #742, opened 2016-11-11
Support Reminder Broadcasts


aaronschachter:
## Tasks
* [x] Add `Broadcast` model
* [x] Refactor Signup Broadcasts to find Campaign based on incoming `broadcast_id` , deprecate `CAMPAIGNBOT_BROADCAST_CAMPAIGN` config variable
* [ ] Add additional check for `broadcast_type` for `reminder` -- If yes response, start/continue Reportback Submission

## Overview

The `chatbot` router currently determines whether an incoming mData request is from a Signup Broadcast by checking for a `req.body.broadcast_id` (#739).

For a Yes response, it will post a Signup to whatever the `CAMPAIGNBOT_BROADCAST_CAMPAIGN` is set for (#718).

This won't work for Reportback Reminder Broadcasts, where we'll be sending multiple Mobile Commons broadcasts for the various CampaignBot Campaigns. We can add support for multiple Mobile Commons Broadcasts by building a new `mobilecommons_broadcasts` collection with fields:

* `broadcast_id`
* `campaign_id`
* `broadcast_type`

When an admin creates a new broadcast in Mobile Commons, they'll need to add the Broadcast ID, the DS Campaign to associate it for, and the broadcast type (`signup` or `reminder`) manually into the `mobilecommons_broadcasts` collection.

When the `chatbot` router receives an mData request containing a `req.body.broadcast_id`, find it in the `mobilecommons_broadcasts` table and take the following action for each `broadcast_type`:

* `signup`
    * Yes response: Signup for the Campaign associated with `broadcast_id`
    * No response: Send CampaignBot `signup_broadcast_declined` message, opt into Agent View

* `reminder`
    * Yes response: Similar to the Reportback start command, either create a new draft and ask for quantity, or continue where member left off in their draft
     * No response: Send new CampaignBot `reminder_broadcast_declined` message, opt into Agent View

cc @jamjensen

jamjensen:
Aaron went on leave right before we made a final decision on how users would respond to Reminder Broadcasts so I went with Papa Roach MVP Last Resort: have members text in campaign keyword.

aaronschachter:
@jamjensen #743 spells out how Reminder Broadcasts + chatbot mData could work per our discussion you and I had either 11/10 or 11/11, and also documents what you just mentioned -- our last resort for reminder messages if we couldn't get Reminder Broadcast mData support in  just asking member to respond with a keyword (no mData support).

aaronschachter:
[...]

PR #718, opened 2016-11-04
Starts on chatbot broadcast query parameter

aaronschachter:
#### What's this PR do?

* Refactors the `loadCampaign` and `loadSignup` Promises in `routes/chatbot` as a single `return` statement

* If a `broadcast` query parameter is set, inspect current User's incoming message as a Yes or No to determine whether to Signup for `CAMPAIGNBOT_BROADCAST_CAMPAIGN`, a new config var to set per broadcast in #651 

    * If user says Yes, signup for the campaign and deliver the `msg_menu_signedup_gambit`

    * If user says No, returns placeholder copy for now and opts the User into Agent View to field any further responses. Does this by throwing a new `Error` for now for sake of simplicity (vs using Promise.cancel, or returning `null` or `false` throughout the Promise chain).

* `routes/chatbot` cleanup:

    *  Use `scope.oip` instead of `mobileCommonsOIP` to DRY

    * Rename `campaignID` as `campaignId` per https://github.com/DoSomething/gambit/pull/697#discussion_r85598802

#### How should this be reviewed?

* Locally can post to `/v1/chatbot?broadcast=true` and respond with according yes/no responses.

* Staging, we'll need to broadcast to our team internally following these steps: 

    * [ ] Create a new CampaignBot Broadcast mData  that posts to Staging `/v1/chatbot?broadcast=true`

    * [ ] Deploy this branch to staging, and set `CAMPAIGNBOT_BROADCAST_CAMPAIGN` to a Campaign ID currently running on CampaignBot

    * [ ] Send broadcast with Conversation set to trigger our new CampaignBot Broadcast mData

* Also ensure sure normal routing of texting in a Campign Signup keyword works 🆗 

#### Any background context you want to provide?

**NOTE**: If user says no to the broadcast, we're still creating a Northstar User for them. We could change this up to just simply return the `'broadcast_declined'` message -- but this seemed to fall into the category of Why not create a User for them. We'll want to clear this with our API and data teams.

Next up:
* [ ] If all looks good, prep a Poduction CampaignBot Broadcast mData posting to prod Gambit `/v1/chatbot?broadcast=true` 
* [ ] Add CampaignBot property to avoid hardcoding "broadcast declined" copy
* [ ] Add a `messages` collection to store incoming `broadcast_id` and whether User signed up or declined the broadcast message they were sent. Refs #698 

#### Relevant tickets
#651 

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.

PR #743, opened 2016-11-12
Adds Broadcast model

aaronschachter:
#### What's this PR do?
* Adds a `broadcasts` collection to associate a Mobile Commons broadcast with its corresponding DS Campaign, broadcast type, and `messages` object
    * Currently only supports a `signup_prompt_declined` message, to display via `mobilecommons.send_message` when User responds with a NO to the Signup broadcast. 
    * Deprecates the CampaignBot `msg_signup_broadcast_declined` message

* Refactors Signup broadcast to load its Campaign by finding a record for the `broadcast_id` in the `broadcasts` table, deprecating the `CAMPAIGNBOT_BROADCAST_CAMPAIGN` config variable (#718)
    * This allows to send multiple Signup broadcasts throughout the week without needing to worry about setting the correct config variable

* Does *NOT* add support for Reminder broadcasts (where the yes/no response should start/continue reportback draft), just gets refactor of Signups working in anticipation of Reminder support

* Adds logging, `MOBILECOMMONS_DISABLED` check to `mobilecommons.send_message` function introduced in #763 

* Add a document into `broadcasts` for a broadcast id, and a CampaignBot campaign X (`type` doesn't really matter because we're not using it yet)
* Post against `chatbot` endpoint passing the `broadcast_id` you added a record for. Confirm a YES response will create the Signup for Campaign X, NO responses returns Campaignbot `signup_broadcast_declined`

* Reminder broadcasts functionality still left to do #742.

#### Relevant tickets
#742 

#### Checklist
- [x] Tested on staging.
- [x] Manually created test`broadcasts` document in staging DB
- [x] Manually created test `broadcasts` document in production DB


app/routes/chatbot.js
```diff
@@ -112,74 +113,82 @@ router.post('/', (req, res) => {
...
+        scope.user = user;
+
+        if (scope.broadcast_id) {
+          return app.locals.db.broadcasts
+            .findById(scope.broadcast_id)
+            .then((broadcast) => {
+              campaign = app.locals.campaigns[broadcast.campaign];
+
+              if (!campaign) {
+                const err = new Error('broadcast campaign undefined');
```
aaronschachter: TODO: Trigger stathat (with email alert) -- broadcast / mData won't work if we don't have a campaign to signup or reportback for.

aaronschachter:
If this branch gets merged we'll also need to update the https://github.com/DoSomething/gambit/wiki/Mobile-Commons#signups documentation to deprecate use of the `CAMPAIGNBOT_BROADCAST_CAMPAIGN` variable and add steps for implementation:
1. Create the desired Mobile Commons Broadcast in Mobile Commons, generating a Broadcast ID
2. Use that ID as the `_id` in the corresponding `broadcasts` document in the Gambit DB, adding the `campaign` we want to Signup for
3. Test by posting to `v1/chatbot`, passing the `broadcast_id` -- verifying the Signup is created for desired Campaign

aaronschachter:
To test this branch, the `broadcasts` collection must be manually created in the Mongo database. I'm adding it into Thor now, and will need to make sure to add it to Production before we plan to send our first broadcast using this model.

One potential improvement to make admin life easier could be to query the Mobile Commons API to get a list of all broadcasts, and automatically create documents for each broadcast. For now, we'll keep it manual and simple for sake of getting this branch live (after being open for oh, 10 weeks, or so)

aaronschachter:
Added a test `broadcasts` document with `_id=123` and `campaign=876` and placeholder `messages.signup_prompt_declined`

Tested on Thor:
* [ ] Passing `broadcast_id=123` and `args=YES` sends CampaignBot Menu Signedup message
* [ ] Continuing conversation submits Reportback successfully
* [ ] Passing `broadcast_id=123` and `args=YES` sends the Broadcast Signup Prompt Declined message
* [ ] Texting a Campaign keyword works as expected
* [ ] Posting a `broadcast_id` that doesn't exist in the `broadcasts` collection results in `{"error":{"code":500,"message":"broadcast undefined"}}`
* [ ] Posting a `broadcast_id` that is not associated with an active CampaignBot campaign results in `{"error":{"code":500,"message":"broadcast campaign undefined"}}`

@sergii-tkachenko This PR is ready for review when you have a chance.

app/models/Broadcast.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Schema.
+ */
+const broadcastSchema = new mongoose.Schema({
+
+  _id: { type: Number, index: true },
+  type: { type: String, enum: ['signup_prompt'] },
+  campaign: Number,
+  messages: {
+    signup_prompt_declined: String,
```
aaronschachter: I'm using the nested `messages` convention that was introduced in #763 -- but I'm not entirely convinced this is the way to go right now since we use our Mongo documents as a CMS. Nesting it as an object makes it a little tricker for our admins to manually edit.

app/models/Broadcast.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+const mongoose = require('mongoose');
+
+/**
+ * Schema.
+ */
+const broadcastSchema = new mongoose.Schema({
+
+  _id: { type: Number, index: true },
+  type: { type: String, enum: ['signup_prompt'] },
```
aaronschachter: Open to other names for this -- eventually there will be another type for `reportback_prompt` per #742

app/models/Broadcast.js
sergii-tkachenko: I talked to @jamjensen earlier about that, and he's okay with it. This do add another one level of complexity, but editing documents in Mongo is and advanced technique by itself.

app/models/Broadcast.js
```diff
@@ -0,0 +1,24 @@
...
+  campaign: Number,
```
sergii-tkachenko: Maybe string? Ids are alphanumeric in Contentful (Phoenix Next): https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/entries/entry/get-a-single-entry

app/routes/chatbot.js
```diff
@@ -112,74 +114,93 @@ router.post('/', (req, res) => {
...
-        const msg = `User ${user._id} current_campaign ${campaignId} not found in CampaignBot.`;
-        throw new NotFoundError(msg);
-      }
+        return resolve(campaign);
+      });
+  });
 
-      return campaign;
-    })
+  return loadCampaign
```
sergii-tkachenko: Semicolon.

app/routes/chatbot.js
```diff
@@ -112,74 +114,93 @@ router.post('/', (req, res) => {
...
+              campaignId = broadcast.campaign;
+              campaign = app.locals.campaigns[campaignId];
+              if (!campaign) {
+                const err = new Error('broadcast campaign undefined');
+                return reject(err);
+              }
+
+              logger.info(`loaded campaign:${campaign._id}`);
+
+              if (campaign.status === 'closed') {
```
sergii-tkachenko: Maybe `!== 'active'` in case we extended campaigns with different states?

app/routes/chatbot.js
```diff
@@ -112,74 +114,93 @@ router.post('/', (req, res) => {
...
+    return loadUser
+      .then((user) => {
+        logger.info(`loaded user:${user._id}`);
+        scope.user = user;
+
+        if (scope.broadcast_id) {
+          return app.locals.db.broadcasts
+            .findById(scope.broadcast_id)
+            .then((broadcast) => {
+              if (!broadcast) {
```
sergii-tkachenko: I'd suggest checking `broadcast._id` instead.

app/routes/chatbot.js
```diff
@@ -112,74 +114,93 @@ router.post('/', (req, res) => {
...
+                // TODO: If this happens, store user/broadcast somewhere to recover lost opportunity
+                // to respond.
+                const err = new Error('broadcast undefined');
```
sergii-tkachenko: Not sure if `UnprocessibleEntityError` makes sense here?

app/routes/chatbot.js
```diff
@@ -112,74 +114,93 @@ router.post('/', (req, res) => {
...
+          campaignId = app.locals.keywords[scope.keyword];
+          campaign = app.locals.campaigns[campaignId];
+
+          if (!campaign) {
+            const msg = `Campaign not found for keyword '${scope.keyword}'.`;
+            const err = new NotFoundError(msg);
+            return reject(err);
+          }
+
+          if (campaign.status === 'closed') {
```
sergii-tkachenko: Same here.

sergiitk (changes requested):

app/models/Broadcast.js
aaronschachter: Let's leave this as number until we're consuming content from a new API and ready to make the change globally -- it's defined as a `Number` in all of the other models.

app/models/Broadcast.js
aaronschachter: Ok, will leave it this way. I'm now leaning toward renaming this as `prompt_declined`, to repurpose the property for the future `reportback_prompt` feature we'll need to close out #742

app/routes/chatbot.js
aaronschachter: I'd prefer to keep this as is until we cross that bridge. If we had statuses like `featured` and `active`, `!==active` would not work.

app/routes/chatbot.js
aaronschachter: Maybe `NotFoundError` ? We can't find a document in the `broadcasts` collection for the `broadcast_id` passed in the request.

app/routes/chatbot.js
aaronschachter: Adding a semi-colon where would break the promise chain below. Would it be better if this line read as 
`return loadCampaign.then((campaign) => {`? The other promises are using the same syntax (where the `then` is on a new line).

app/routes/chatbot.js
sergii-tkachenko: Yes, makes more sense.

app/routes/chatbot.js
sergii-tkachenko: Ah, yes, I missed the promise below. Totally my fault.

app/routes/chatbot.js
```diff
@@ -112,74 +114,91 @@ router.post('/', (req, res) => {
...
+    let campaign;
+    let campaignId;
+
+    return loadUser
+      .then((user) => {
+        logger.info(`loaded user:${user._id}`);
+        scope.user = user;
+
+        if (scope.broadcast_id) {
+          return app.locals.db.broadcasts
```
sergii-tkachenko: Would it make sense to move this to a helper function, something like `loadBroadcastCampaign` to reduce nesting?

sergiitk (changes requested):

app/routes/chatbot.js
```diff
@@ -271,8 +290,20 @@ router.post('/', (req, res) => {
...
+          const logMsg = 'undefined broadcast.messages.prompt_declined';
+          logger.error(logMsg);
+          stathat(`error: ${logMsg}`);
+
+          // Don't send an error code to Mobile Commons, which would trigger error message to User.
+          return helpers.sendResponse(res, 200, logMsg);
+        }
+
+        // Log the no response:
+        app.locals.db.bot_requests.log(req, 'broadcast', null, 'prompt_declined', msg);
```
aaronschachter: Seems slightly hacky to pass `broadcast` as `bot_type` and `null` as the `bot_id` -- but we're defining the `prompt_declined` copy on the `broadcast` model. We already are storing the `broadcast_id` within in the `body` object in the model, which is how we currently query the collection to determine YES responses.

screen shot 2017-01-26 at 8 22 04 am

app/routes/chatbot.js
aaronschachter: It sounded like it made sense, but realized that we still need the loaded broadcast model [in this line](https://github.com/DoSomething/gambit/blob/f71df68b0b64f19588bd19cd85e660c348cfa2b1/app/routes/chatbot.js#L153) to pass the Broadcast `prompt_declined` message later. Adding a static `loadBroadcast` method to the `broadcast` model would still look the same.

sergii-tkachenko: Let's discuss it tomorrow anyway.

aaronschachter:
@jamjensen @classicfreddie This was deployed to prod today ([3.3.0](https://github.com/DoSomething/gambit/releases/tag/3.3.0)). I've removed the `CAMPAIGNBOT_BROADCAST_CAMPAIGN` from Heroku and updated the docs: https://github.com/DoSomething/gambit/wiki/Chatbot#signup-broadcasts

PR #748, opened 2016-11-16
Yes response refactoring

### LINKED ISSUE #717: Update logic for Yes response
Current `isYesResponse` function (that we used for "Can you send photos from your phone?" legacy `/yes-no-path` endpoint) [works this way:](https://github.com/DoSomething/gambit/blob/031b37ace0fdb650ac516aabc9b3d9b19cd714ab/lib/helpers.js#L163
)

If the first word isn't  'N', 'NA', 'NAH', 'NO', or 'NIL' (not case sensitive)-- consider the response a Yes. This added in #598 was to handle #513 -- adding support for different Yes variations like "sure", "of course", "ya", etc

Need product team to confirm logic or specify requirements. If we need to change current behavior, then this issue can remain open until we implement the changes in code.

Refs #651

deadlybutter:
#### What's this PR do?
- Checks if the word is in a list of "yes'es", "yes's", i dont know one words this.
- Simplifies the code & removes unnecessary logic, variables

- I ran the function through the node console because its so simple
screen shot 2016-11-16 at 11 10 29 am

nope

#### Relevant tickets
Fixes #717


lib/helpers.js
```diff
@@ -53,22 +53,10 @@ module.exports.getFirstWord = function (message) {
...
-    'NAH',
-    'NO',
-    'NIL',
-  ];
-
-  if (noResponses.indexOf(trimmed) >= 0) {
-    return false;
-  }
+  const trimmed = this.getFirstWord(message.toLowerCase().trim());
+  const yesResponses = process.env.GAMBIT_YES_RESPONSES || ['yes'];
```
deadlybutter: actually, i should probably convert the string into an array :facepalm:

deadlybutter:
updated

screen shot 2016-11-16 at 11 26 52 am

lib/helpers.js
```diff
@@ -51,24 +51,11 @@ module.exports.getFirstWord = function (message) {
...
-    'NAH',
-    'NO',
-    'NIL',
-  ];
-
-  if (noResponses.indexOf(trimmed) >= 0) {
-    return false;
-  }
+  const trimmed = this.getFirstWord(message.toLowerCase().trim());
+  const yesResponses = process.env.GAMBIT_YES_RESPONSES ? `${process.env.GAMBIT_YES_RESPONSES}`.split(',') : ['yes'];
```
sergii-tkachenko: Why do you need to convert `process.env.GAMBIT_YES_RESPONSES` to a string, isn't it a string already?

sergii-tkachenko:
Check this out: https://www.npmjs.com/package/yesno

sergii-tkachenko:
Even better: https://www.npmjs.com/package/yn

lib/helpers.js
deadlybutter: being overly cautious I guess, yeah probably ok to just call .split

deadlybutter:
I'll checkout the packages now!

deadlybutter:
Hmm, the yn one seems like the better option but it feels overkill to install a package in order to replace two lines of code

sergii-tkachenko:
Well, it's not 2 lines: https://github.com/sindresorhus/yn/blob/master/index.js.
I still like `yn` package:
- We won't need to introduce another one environment variable. I see no reasons why we need to move it to an environment, it's not like it's differs from local to prod. It's not a secret key eather
- It's already canse-insensetive
- It has lenient mode to process typos if we want to: https://github.com/sindresorhus/yn/blob/master/lenient.js

deadlybutter:
Oh I didnt see the scoring system, that's really useful : thumbsup:

deadlybutter:
@sergii-tkachenko wait, we cant use this actually. james has a list of yes responses that is far more extensive then what this package seems to offer

deadlybutter:
For reference this is everything we want supported https://github.com/DoSomething/gambit/issues/717#issuecomment-259281947

sergii-tkachenko:
Makes sense to use own code then.

deadlybutter:
Good to merge?

lib/helpers.js
```diff
@@ -51,24 +51,12 @@ module.exports.getFirstWord = function (message) {
...
-    'NO',
-    'NIL',
-  ];
-
-  if (noResponses.indexOf(trimmed) >= 0) {
-    return false;
-  }
+  const trimmed = this.getFirstWord(message.toLowerCase().trim());
+  const yesResponses = process.env.GAMBIT_YES_RESPONSES ?
+    `${process.env.GAMBIT_YES_RESPONSES}`.split(',') : ['yes'];
```
sergii-tkachenko: Should you just call split on `process.env.GAMBIT_YES_RESPONSES`?

deadlybutter: Ah right, forgot to fix that one

sergii-tkachenko: Also, please lowercase yes responses.

deadlybutter:
@sergii-tkachenko Lowercased everything, also took care of removing the unneeded string

lib/helpers.js
```diff
@@ -51,24 +51,12 @@ module.exports.getFirstWord = function (message) {
...
-    'NO',
-    'NIL',
-  ];
-
-  if (noResponses.indexOf(trimmed) >= 0) {
-    return false;
-  }
+  const trimmed = this.getFirstWord(message.toLowerCase().trim());
+  const yesResponses = process.env.GAMBIT_YES_RESPONSES ?
+    process.env.GAMBIT_YES_RESPONSES.split(',').map(res => res.toLowerCase()) : ['yes'];
```
sergii-tkachenko: Maybe just lowercase it while it's still a string?

deadlybutter: Yep, that seems simpler

sergii-tkachenko:
One last thing, could you please add `GAMBIT_YES_RESPONSES` to `.env.example`?

deadlybutter:
Sure thing! I left that default array just in case someone forgets but good to have it documented in there

sergii-tkachenko:
Thank you!

Issue #704, opened 2016-10-31
Investigate response times


aaronschachter:
Heroku has been sending a number of response time alerts, which seem to have started the [first day we pushed the CampaignBot](https://github.com/DoSomething/gambit/releases/tag/2.5.1) code to production 😿

aaronschachter:
Knocking on wood, but things seem quieter today. As of this morning, we no longer have mData's posting to the legacy `reportback/:campaignName` endpoint (#706) -- and sending members only to the `chatbot` endpoint for Campaigns.

deadlybutter:
Looking over Heroku metrics now,

Quite a few events are closely tied to other actions we've taken,
<img width="372" alt="screen shot 2016-11-02 at 4 39 03 pm" src="https://cloud.githubusercontent.com/assets/897368/19946632/176edd70-a11b-11e6-90e4-c508a9967464.png">

But quite a few aren't (Including the two errors fired this afternoon)
<img width="457" alt="screen shot 2016-11-02 at 4 41 05 pm" src="https://cloud.githubusercontent.com/assets/897368/19946656/2d294952-a11b-11e6-8ca3-ab54a37be96c.png">

It also seems unrelated to throughput or cpu load
<img width="400" alt="screen shot 2016-11-02 at 4 42 50 pm" src="https://cloud.githubusercontent.com/assets/897368/19946722/8361220e-a11b-11e6-9d60-12077560df9f.png">
(notice the throughput on the bottom has bars in places outside the error which are much higher)

Tomorrow I'm going to do some thorough profiling, these are the steps ill take (in my head)

* Create a test dyno formation w/ a custom version of Gambit. In the procfile define the command to start as `node --prof server.js` instead of the current value.
* Generate substantial number of [requests to chatbot](https://github.com/DoSomething/gambit/blob/develop/documentation/endpoints/chatbot.md#chat.
* Pull profile logs from server w/ `heroku run bash --app ${testservername}`
* Process logs into something readable locally `node --prof-process isolate-0xnnnnnnnnnnnn-v8.log > processed.txt`

Hopefully this nets some valuable insight.

More info here from NodeJS docs: 
https://nodejs.org/en/docs/guides/simple-profiling/

aaronschachter:
Thanks @deadlybutter -- looking forward to seeing the results. 

[...]

PR #752, opened 2016-11-18
Properly handling multiple errors in the signup endpoint

deadlybutter:
#### What's this PR do?
- We're getting a lot of errors being reported at times because they are not being caught correctly, this addresses that.

I simply put a throw at the top of the endpoint

screen shot 2016-11-18 at 10 58 12 am

And hit the endpoint twice (once for each error type)
screen shot 2016-11-18 at 10 58 03 am
screen shot 2016-11-18 at 10 58 36 am

```
Joe Kent	[10:25 AM]  
the errors being thrown are `HttpError 422 - Unprocessable Entity` on the /v1/signups endpoint, these seem to be thrown here 
https://github.com/DoSomething/gambit/blob/f514f4113c19f6dcd80828f0281974b1a0365e17/app/routes/signups.js#L43

the problem is the first catch statement isnt checking for that type of error, https://github.com/DoSomething/gambit/blob/f514f4113c19f6dcd80828f0281974b1a0365e17/app/routes/signups.js#L80
and because the error isnt passed down those other catches (including the one for UnprocessibleEntityError) are never executed, and thus never send a response (edited)
```
from [Slack](https://dosomething.slack.com/archives/gambit/p1479482735000008)

#### Relevant tickets
Semi related to #704 in that it relates to stability & monitoring


app/routes/signups.js
```diff
@@ -77,10 +77,13 @@ router.post('/', (req, res) => {
 
       return helpers.sendResponse(res, 200, scope.response_message);
     })
-    .catch(NotFoundError, err => helpers.sendResponse(res, 404, err.message))
-    .catch(UnprocessibleEntityError, err => helpers.sendResponse(res, 422, err.message))
-    .catch(err => helpers.sendResponse(res, 500, err.message));
+    .catch(err => {
```
sergii-tkachenko: Add parentheses around `err`.

app/routes/signups.js
```diff
@@ -77,10 +77,13 @@ router.post('/', (req, res) => {
...
+      switch (err.name) {
```
sergii-tkachenko: There's a better way to do it. Import bluebird and catch them separately: http://bluebirdjs.com/docs/api/catch.html#filtered-catch

image

app/routes/signups.js
```diff
@@ -5,6 +5,7 @@ const router = express.Router(); // eslint-disable-line new-cap
 const helpers = require('../../lib/helpers');
 const NotFoundError = require('../exceptions/NotFoundError');
 const UnprocessibleEntityError = require('../exceptions/UnprocessibleEntityError');
+const Promise = require('bluebird'); // eslint-disable-line no-unused-vars
```
sergii-tkachenko: Could you please add a comment why we need to import bluebird here? I just worry that it might get deleted someday, on incorrect assumption that it's not used.

PR #753, opened 2016-11-20
Adds Docker support to Gambit

deadlybutter:
#### What's this PR do?
- Updates winston-mongodb to latest version (Had NPM error blocking startup)
- Updates Procfile command
- Adds Dockerfile & dockerignore

This repo itself should be build-able. However to run it you'll need to use the Docker Compose file [here](https://github.com/DoSomething/gambit-services/blob/master/docker-compose.yml)

In order to test the build process still works on Heroku I made a deploy with this branch to staging,
- The build log is here (Looks ✅ ) https://dashboard.heroku.com/apps/ds-mdata-responder-staging/activity/builds/ef1ca1a8-63e6-4bbc-822a-c71686d3b693
- The logs following the app start also seem good
screen shot 2016-11-20 at 1 16 13 pm

As we start moving towards more services, docker compose will make it much easier for developers to get up & running with Gambit using a consistent image across all developers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM node:6.9.1
+
+RUN npm install -g foreman
+
+# Create app directory
+RUN mkdir -p /app
+WORKDIR /app
+
+# Install app dependencies
+COPY package.json /app
+RUN npm install
+
+# Bundle app source
+COPY . /app
+
+CMD nf start
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,12 @@
 FROM node:6.9.1
 
-# Packages required to run the app
-RUN npm install -g foreman
-RUN npm install -g nodemon
-
 # Create app directory
-RUN mkdir -p /app
 WORKDIR /app
-
-# Install app dependencies
-COPY package.json /app
-RUN npm install
 
 # Bundle app source
 COPY . /app
+
+# Get NPM Modules
+RUN npm install && npm install -g foreman nodemon
 
 CMD nodemon --exec "nf start"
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM node:6.9.1
+
+# Packages required to run the app
+RUN npm install -g foreman
```
sergii-tkachenko: The less RUN commands there are, the better. Every command creates another one fs layer with its own uuid. Take a look how docker files are structured for most packages:

https://github.com/nodejs/docker-node/blob/d20d305f0bf5935385a32558501f3a5c65e34878/6.9/Dockerfile#L23
https://github.com/docker-library/php/blob/b208b3bd750feea2c236af0586e544dd1daad8a2/7.0/Dockerfile#L41

They have as many commands separated by semicolon in one RUN as possible.

deadlybutter: @sergii-tkachenko Oh interesting, didn't know that! Let me reduce that down then

README.md
```diff
@@ -1,2 +1,24 @@
...
+1. `git clone`
+2. `docker-compose up`
+
+### Docker setup under the hood
+All apps are executed by `Foreman` to handle process management & mimic Heroku.
+`Nodemon` will autoreload the server when a file changes.
+The compose file defines env variables for connection details & network mapping.
+
+## Standalone without Docker
+1. Run an instance of RabbitMQ, MongoDB. Two options,
```
aaronschachter: @deadlybutter We're not using RabbitMQ for anything, right?  I'm looking to remove this "Standalone" section since it doesn't seem accurate. 

There's could be argument to remove these changes entirely if you won't be able to help out with support...

deadlybutter: @aaronschachter I think in another issue i said it's totally cool to remove this if you're not going to use it!

PR #755, opened 2016-11-24
Adds node modules as a volume

deadlybutter:
#### What's this PR do?
Mounts a volume for node_modules, apparently its that easy!
http://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html
http://stackoverflow.com/questions/30043872/docker-compose-node-modules-not-present-in-a-volume-after-npm-install-succeeds

#### How should this be reviewed?
Can you docker-compose up without running npm install?

#### Relevant tickets
Fixes #754

PR #756, opened 2016-11-25
Activity Events

### LINKED ISSUE #754: Figure out optimal Dockerfile
Current setup uses a `volume` to transfer code over into the container

So when we run npm install on build, the modules get installed. But the volume then overrides the `node_modules` folder with nothing.

The temp solution for now is to run `npm install` locally.

Ideally we should not require this step.

deadlybutter:
#### What's this PR do?
- Produces `signup` and `reportback` events to Rabbit MQ  `activity events` queue
- Updates docker compose setup to add RabbitMQ container
 - Also includes changes from https://github.com/DoSomething/gambit/pull/755 so I wasn't blocked from working.

Post to Gambit locally & check the events are fired

Needed for relative reminder messaging

#### Relevant tickets
Fixes #754 
Touches #742

### DIFF
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -6,6 +6,7 @@
 const logger = app.locals.logger;
 const campaignBot = app.locals.campaignBot;
 const helpers = rootRequire('lib/helpers');
+const rabbit = app.locals.rabbit;
 
 /**
  * CampaignBotController.
@@ -118,6 +119,13 @@ class CampaignBotController {
   createReportbackSubmission(req) {
     this.debug(req, 'createReportbackSubmission');
 
+    rabbit.produce('activity events', {
+      type: 'signup',
+      user: req.user,
+      campaignId: req.campaign._id,
+      campaignRunId: req.campaign.current_run,
+    });
+
     return req.signup
       .createDraftReportbackSubmission()
       .then(() => {
@@ -178,6 +186,13 @@ class CampaignBotController {
    */
   postReportback(req) {
     this.debug(req, 'postReportback');
+
+    rabbit.produce('activity events', {
+      type: 'reportback',
+      user: req.user,
+      campaignId: req.campaign._id,
+      campaignRunId: req.campaign.current_run,
+    });
 
     return req.signup
       .postDraftReportbackSubmission()
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -6,7 +6,6 @@
 const logger = app.locals.logger;
 const campaignBot = app.locals.campaignBot;
 const helpers = rootRequire('lib/helpers');
-const rabbit = rootRequire('lib/rabbit');
 
 /**
  * CampaignBotController.
@@ -179,13 +178,6 @@ class CampaignBotController {
    */
   postReportback(req) {
     this.debug(req, 'postReportback');
-
-    rabbit.produce('activity events', {
-      type: 'reportback',
-      userId: req.user._id,
-      campaignId: req.campaign._id,
-      campaignRunId: req.campaign.current_run,
-    });
 
     return req.signup
       .postDraftReportbackSubmission()
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -6,7 +6,7 @@
 const logger = app.locals.logger;
 const campaignBot = app.locals.campaignBot;
 const helpers = rootRequire('lib/helpers');
-const rabbit = app.locals.rabbit;
+const rabbit = rootRequire('lib/rabbit');
 
 /**
  * CampaignBotController.
@@ -119,13 +119,6 @@ class CampaignBotController {
   createReportbackSubmission(req) {
     this.debug(req, 'createReportbackSubmission');
 
-    rabbit.produce('activity events', {
-      type: 'signup',
-      user: req.user,
-      campaignId: req.campaign._id,
-      campaignRunId: req.campaign.current_run,
-    });
-
     return req.signup
       .createDraftReportbackSubmission()
       .then(() => {
@@ -189,7 +182,7 @@ class CampaignBotController {
 
     rabbit.produce('activity events', {
       type: 'reportback',
-      user: req.user,
+      userId: req.user._id,
       campaignId: req.campaign._id,
       campaignRunId: req.campaign.current_run,
     });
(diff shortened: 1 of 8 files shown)


lib/rabbit.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function makeConnection() {
+  logger.info('Attempting amqp connection');
+
+  const amqp = require('amqplib');
+
+  amqp.connect(process.env.RABBIT_URI)
+  .then(connection => connection.createChannel())
+  .then(ch => {
```
sergii-tkachenko: Should be a oneliner.

lib/rabbit.js
```diff
@@ -0,0 +1,39 @@
...
+    channel = ch;
+  })
+  .catch(err => {
```
sergii-tkachenko: Parentheses.

lib/rabbit.js
```diff
@@ -0,0 +1,39 @@
...
+    onError(err);
+    setTimeout(makeConnection, 2000);
```
sergii-tkachenko: Why 2000?

lib/rabbit.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+}
+makeConnection();
+
+module.exports = {
+  produce(queueName, message) {
+    const outgoing = JSON.stringify(message);
+
+    return channel
```
sergii-tkachenko: Not readable.

lib/rabbit.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const stathat = app.locals.stathat;
+const logger = app.locals.logger;
+
+let channel = {};
+
+function onError(err) {
+  logger.error('Rabbit Error', err);
+  stathat(`gambit-core rabbit error ${err.code}`);
```
sergii-tkachenko: What if it's not an exception? You need to handle err.code = undefined.

lib/rabbit.js
sergii-tkachenko: Also, `ch` is not informative.

lib/rabbit.js
```diff
@@ -0,0 +1,39 @@
...
+  .then(connection => connection.createChannel())
+  .then(ch => {
+    channel = ch;
+  })
+  .catch(err => {
+    onError(err);
+    setTimeout(makeConnection, 2000);
+  });
+}
+makeConnection();
```
sergii-tkachenko: This is bad. Really difficult to notice. Make this a class with a constructor.

lib/rabbit.js
deadlybutter: Yeah it was a bit hard to name this because I couldn't use `channel`. Suggestions?

lib/rabbit.js
deadlybutter: No specific reason, just felt like 2 seconds was a reasonable retry timeout

lib/rabbit.js
deadlybutter: What is not readable?

lib/rabbit.js
sergii-tkachenko: Refactor this to a class. So channel will be `this.channel`.

lib/rabbit.js
sergii-tkachenko: This should be in a constant then.

deadlybutter:
updated but not tested, also seems to have lint problem

lib/rabbit.js
```diff
@@ -0,0 +1,46 @@
...
+      this.channel = channel;
+    })
+    .catch((err) => {
+      onError(err);
+      setTimeout(this.connect, AMQP_RETRY_FREQUENCY);
+    });
+  }
+}
+
+const rabbitConnection = new Connection();
```
sergii-tkachenko: I meant you can just put everything in the class and export it.

app/controllers/CampaignBotController.js
```diff
@@ -179,6 +180,13 @@ class CampaignBotController {
   postReportback(req) {
     this.debug(req, 'postReportback');
 
+    rabbit.produce('activity events', {
+      type: 'reportback',
+      userId: req.user._id,
+      campaignId: req.campaign._id,
+      campaignRunId: req.campaign.current_run,
+    });
```
deadlybutter: @aaronschachter I did not see that issue, thanks for pointing that out! I'll move this over to Signup then

deadlybutter:
@sergii-tkachenko sweet, i think this is much better now

aaronschachter:
@deadlybutter I'm not following the logic in this branch, and there's no corresponding issue to explain what's going on here. Why are we posting `signup` and `reportback` events?

deadlybutter:
@aaronschachter By posting events to a message broker we can more easily spin up new micro services (in contrast to adding more uri/keys everywhere & relying on HTTP). In this case, the next step is a micro service to keep track of sending relative reminder messages. It will consume these activity events to figure out when it should be posting reminders. Sorry I haven't documented any of this anywhere yet in Github or Trello, all over the place atm but I will make sure to get that done for you!

aaronschachter:
We're posting Signups and Reportbacks to the Phoenix API -- which posts to Quicksilver, so isn't posting events to a message broker already taken care of?  Is the intent to implement something like https://github.com/DoSomething/quicksilver-api/issues/98?

I know I'm on leave, so take this with a grain of salt -- but my $0.02 on this one is to hold off on any more development (and hold on merging) until there's a full technical spec that documents what the intent of this work is for, and how it will be implemented.

deadlybutter:
@sergii-tkachenko Im going to reopen the other PR that had the Docker fixes from this PR, as those should still be merged in anyway

@aaronschachter ok, update! I talked w/ Sergii for a bit -- we're both still onboard with _moving_ Gambit towards a more distributed system versus a single application. I think we've talked about this before but I would be happy to write out the reasoning if im mistaken!

However we both agreed to minimize the scope as much as possible. For now, since Quicksilver already has signup/reportback events, we're just going to tap into that system. We're going to add a signup/reportback reminder message to the campaign object & expose that in the API. From there Quicksilver can use that to send out reminders. Let me know if you have any more questions!

aaronschachter:
@deadlybutter I do think we need at least a Github issue to serve as the spec for why distributed system and specifically when/where we'll use it.

Using Quicksilver for sending relative reminders definitely makes sense -- @sergii-tkachenko and I had mentioned it either should belong in QS or the source system of Signups (which would be Phoenix... but eventually Rogue.. so definitely makes sense for QS to handle to unblock development)

PR #757, opened 2016-11-28
Adds broadcast ID to signup object

deadlybutter:
#### What's this PR do?
Adds broadcast ID to signup object.

This will let us query for total broadcast signups & reportbacks really easily.

Signup for a campaign via broadcast

#### Relevant tickets
Fixes #N/A

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -208,7 +208,7 @@ router.post('/', (req, res) => {
 
       if (scope.broadcast_id) {
         scope.signup.broadcast_id = scope.broadcast_id;
-        scope.signup.save();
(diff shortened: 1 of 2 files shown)


app/routes/chatbot.js
```diff
@@ -206,6 +206,11 @@ router.post('/', (req, res) => {
         return false;
       }
 
+      if (scope.broadcast_id) {
+        scope.signup.broadcast_id = scope.broadcast_id;
+        scope.signup.save();
```
sergii-tkachenko: Then/catch it here and make sure it you log the error message on failure.

deadlybutter:
[Query for total signups & rb's from a broadcast](https://trello.com/c/6R646aEE/1-query-for-total-signups-rb-s-from-a-broadcast)

PR #763, opened 2016-11-30
Adds support for external reminder notifications

deadlybutter:
#### What's this PR do?
Adds an endpoint for triggering reminder notifications 

#### How should this be reviewed?
- hit endpoint w/ missing params 
- hit endpoint w/ all params but missing reminder on campaign
- hit endpoint w/ all params & set reminder msg on campaign

screenshot_20161130-144245

#### Relevant tickets
Fixes #760

_Still need to update docs for this, not ready for merge_

Issue #749, opened 2016-11-16
422 error on /v1/signup


sergii-tkachenko:
https://rpm.newrelic.com/accounts/108038/applications/25522350/traced_errors/41728349-ac56-11e6-b5b0-b82a72d2466d_5765_6589

2016-11-16T23:27:08.282576+00:00 app[web.1]: error: Missing required user.mobile.
2016-11-16T23:27:08.251024+00:00 heroku[router]: at=info method=POST path="/v1/signups/" host=ds-mdata-responder.herokuapp.com request_id=e9f8eb0d-ea84-4780-b9df-60739f927c17 fwd="52.3.68.224" dyno=web.1 connect=0ms service=183ms status=422 bytes=284

aaronschachter:
@sergii-tkachenko This error is expected when Quicksilver posts a Signup and the corresponding Northstar user doesn't have a mobile number set -- there's no number to deliver the confirmation message to

sergii-tkachenko:
Yes, in hindsight I should've created this one in Quicksilver-PHP to investigate how it passed messages with no phones in them.

aaronschachter:
@sergii-tkachenko Does it make sense to close this? I know you're working on this in Quicksilver... but I don't think we'll be making any Gambit changes to resolve...

sergii-tkachenko:
👍

Issue #635, opened 2016-09-28
Investigate polling for messages and responding by POST send_message


aaronschachter:
This could be a big performance win. We currently game the Mobile Commons system by posting to the `profile_update` endpoint, saving our chatbot response to a Custom Field (`gambit_chatbot_response`) and then opting the member into an OIP that renders the `{{gambit_chatbot_response}}`` Custom Field value via Liquid.

Per Travis:

> That would do a mail merge on the message body, which adds time to message generation since we have to lookup the custom field as well as constantly updating a user’s profile.

We could potentially poll the `messages` GET endpoint to inspect using the `phone_number` and `body` of incoming messages. Figure out what to respond to each user with and post to the `send_message` endpoint. Repeat.

This could fix a nasty bug we get where a member sends us 3 texts us in a row before receiving a message back from the 1st text send. Gambit processes the 2nd text sent but it may not be the answer to our 1st response back.

cc @sergii-tkachenko

aaronschachter:
@jamjensen Today in Story Time ™️  📚 🔮 , we were discussing support for attaching an image to a message we send through CampaignBot, e.g. including a sample Reportback image in a Signup confirmation message. 

This issue is related to that feature request -- if we refactored CampaignBot to deliver messages by posting to the Mobile Commons `send_message`endpoint, we'd be able to include an attachment in our post data. We don't have the ability to do that via current approach (we send SMS messages by posting to the Mobile Commons `profile_update` endpoint and displaying our `{{gambit_chatbot_response}}` as Liquid in the OIP we opt member into)

cc @justkika @classicfreddie

aaronschachter:
Closing this for now until approved/scheduled for development -- added it to our [Feature Request doc](https://docs.google.com/document/d/1LwCUbm81KriTb8pQtRDI8RkX8DLF-1qR9CjfUT3x6DA)

aaronschachter:
Wanted to document that since our end goal is switching processing incoming SMS messages to a Twilio webhook -- this issue is no longer relevant. We'd want to keep the Gambit setup as is: posting all incoming messages to the `chatbot` endpoint. The main change would be how we respond back: using the Twilio API instead of our current Mobile Commons `profile_update` / Opt-in path implementation.

PR #766, opened 2016-12-02
Reminders: change input param to mobile, add logging.

sergiitk:
#### What's this PR do?
- Adds more checks and logging to relative reminders
- Changes input params format from camelCase to snake_case, as we do [everywhere in Northstar](https://github.com/DoSomething/northstar/tree/dev/documentation)
- Changes required `northstarUserId` to `mobile`, as quicksilver doesn't always know the id
- Significantly improves logging

```
curl -H "x-gambit-api-key: totallysecret" -H "Accept: application/json" -H "Content-Type: application/json" -d '{"mobile": "3478227222", "campaign_id": "4944", "reminder_type": "signup"}' http://localhost:5000/v1/notifications/reminder
{"success":true}
```

#### Relevant tickets
References:
- #763 
- https://github.com/DoSomething/Quicksilver-PHP/issues/78
- Trello: https://trello.com/c/82XOQRkv/134-quicksivler-as-a-member-i-want-to-receive-a-reminder-message-relative-to-when-i-sign-up

#### Checklist
- [x] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love.


sergii-tkachenko:
One more question: is `/notifications/` NS set in stone?

aaronschachter:
I think there's a good argument to rename the `notifications` endpoint, esp if we implement something like #635 to implement conversation without using mData's and Opt-in paths.

Also glad to see renaming `northstarUserId`. But if the given mobile number doesn't have a Northstar ID, don't we run into issues where all future conversations won't work since each `chatbot` request finds Northstar ID based on given mobile number? refs https://github.com/DoSomething/gambit/issues/749#issuecomment-261112025

app/routes/notifications.js
```diff
@@ -2,41 +2,64 @@
...
-router.post('/reminder', (req, res) => {
-  const northstarUserId = req.body.northstarUserId;
-  const campaignId = req.body.campaignId;
-  const reminderType = req.body.reminderType;
+const allowedReminderTypes = ['signup', 'reportback'];
 
-  if (!northstarUserId || !campaignId || !reminderType) {
-    res.json({ error: 'Missing parameters' });
-    return;
+router.post('/reminder', (req, res) => {
```
aaronschachter: Refs https://github.com/DoSomething/gambit/pull/766#issuecomment-264667391 -- if Quicksilver was inspecting all incoming messages for conversation, we could just post to an endpoint like `send_message` and send along the CampaignBot message type as a parameter in our POST, instead of having different routes for each Bot message:
e.g. 
* POST `send_message` mobile=2125551234 message_type='reminder' campaign=362
* POST `send_message` mobile=2125557788 message_type='ask_quantity' campaign=2070

deadlybutter:
@aaronschachter I would rather not expose access to send_sms (or send_fb_message, etc) via a Gambit API -- I think thats a really significant change we'd need to think about in terms of security & product (eg: If Rogue wants to hook into this, what is the process like? Do they ask us for copy, to review it, etc). For now I would prefer to only expose messaging we've created & decided is ok to be sent by 3rd party apps.

app/routes/notifications.js
```diff
@@ -2,41 +2,64 @@
...
+    return helpers.sendResponse(res, 422, 'Missing required campaign id parameter.');
+  }
+  if (!req.body.reminder_type) {
+    return helpers.sendResponse(res, 422, 'Missing required reminder type parameter.');
   }
 
-  if (!app.locals.campaigns[campaignId]) {
-    res.json({ error: 'No campaign set in Gambit for the given campaign id.' });
-    return;
+  const mobile = req.body.mobile;
```
deadlybutter: I think it would be simpler to just define all of these variables first, then check if they are defined

```js
const mobile = req.body.mobile;

if (!mobile) { ... }
```

otherwise you're typing out `req.body.<something>` twice

itsjoekent (changes requested):
Small comment here but otherwise the rest looks good!

aaronschachter:
@deadlybutter My mock requests in https://github.com/DoSomething/gambit/pull/766#discussion_r90765997 aren't passing the raw message copy to deliver to an end user, it's passing the Bot message type to send (`msg_reminder_24hrs`, `msg_ask_quantity`).  So maybe a better namespace choice could be `send_bot_message` instead of `send_message`, where we'd likely want to pass a [`bot_type` too](https://github.com/DoSomething/gambit/blob/fbe1ead5067a50f137f1573cb5954e8efb212a0f/app/models/BotRequest.js#L16) if/when we add additional bots (Onboarding, SMS Game, Account Management, etc)

app/routes/notifications.js
sergii-tkachenko: Good idea!

sergii-tkachenko:
I agree with Aaron here. Except there's no `bots` table for now, or even a list of available bots. So I'll just implement `send_bot_message` without `bot_type` argument.

app/routes/notifications.js
```diff
@@ -2,41 +2,64 @@
...
 const express = require('express');
 const router = express.Router(); // eslint-disable-line new-cap
+const helpers = require('../../lib/helpers');
 const mobilecommons = rootRequire('lib/mobilecommons');
 const logger = app.locals.logger;
 const stathat = app.locals.stathat;
 
+const allowedReminderTypes = ['signup', 'reportback'];
+
 router.post('/reminder', (req, res) => {
```
aaronschachter: @sergii-tkachenko Were you going to change this `/notifications/reminder` endpoint, or is that for a different PR?

app/routes/notifications.js
```diff
@@ -2,41 +2,64 @@
 
 const express = require('express');
 const router = express.Router(); // eslint-disable-line new-cap
+const helpers = require('../../lib/helpers');
 const mobilecommons = rootRequire('lib/mobilecommons');
 const logger = app.locals.logger;
 const stathat = app.locals.stathat;
 
+const allowedReminderTypes = ['signup', 'reportback'];
```
deadlybutter: No we send it for both

app/routes/notifications.js
sergii-tkachenko: I'll change it in the next commit.

sergii-tkachenko:
I've decided to make it a [REST action](http://restful-api-design.readthedocs.io/en/latest/methods.html#actions) on campaigns. Here's how the request look like:

```
curl -H "x-gambit-api-key: totallysecret" -H "Accept: application/json" -H "Content-Type: application/json" -d '{"mobile": "3478227222", "type": "signup"}' http://localhost:5000/v1/campaigns/4944/message
{"success":true}
```


Please re-review.

sergii-tkachenko:
I deliberately haven't updated the docs, will do it when all decisions are confirmed.

app/models/Campaign.js
```diff
@@ -34,8 +34,12 @@ const campaignSchema = new mongoose.Schema({
   msg_menu_signedup_external: String,
   msg_menu_signedup_gambit: String,
   msg_no_photo_sent: String,
-  msg_relative_reminder_signup: String,
-  msg_relative_reminder_reportback: String,
+
+  // Exposed messages.
+  messages: {
```
deadlybutter: Eventual refactor: Lets nest all of the messages instead of prefixing `msg_`

deadlybutter: Also, should this be `signup_reminder`? Im just worried `signup` could be confused with a "Thanks for signing up" or something similar

deadlybutter: Or maybe we do 
```
reminder_messages: {
  signup:
  reportback:
}

**Review by itsjoekent — CHANGES_REQUESTED** [review]
Some thoughts on naming, otherwise great!

File: app/models/Campaign.js (hunk 5 above)
Comment (aaronschachter) [inline app/models/Campaign.js]: Is this an actual "reminder" to signup (suggesting user has already gotten a message to signup)?

Or just the first broadcast message sent prompting a user to signup for a new campaign? (Where using the term "reminder" is misleading)

Comment (sergii-tkachenko) [inline app/models/Campaign.js]: Signup message is intended to remind a user to report back next day after signup event at 3PM.
Reportback message is intended to encourage a user to sign up for a new campaign after they reported back event in a few days after.

Comment (sergii-tkachenko) [inline app/models/Campaign.js]: We've discussed the wording issue in the team and have agreed to name first `relative_to_signup` and second `relative_to_reportback`. Let me know if it you're ok with it @aaronschachter.

Comment (aaronschachter) [inline app/models/Campaign.js]: I think that's reversed @sergii-tkachenko -- but I follow. It's semantics but Signup Reminder message sounds ;ike "Hey you haven't signed up for this yet" -- i don't think it should be called a reminder message.

**Comment by sergii-tkachenko** [comment]
Is it goo to go?

File: app/models/Campaign.js (hunk 5 above)
Comment (deadlybutter) [inline app/models/Campaign.js]: Its a message reminding you to do something, relative to another event -- Im not sure of a better naming system?

File: app/routes/campaigns.js [hunk 6]
```diff
@@ -41,4 +46,51 @@ router.get('/:id', (req, res) => {
...
+    const msg = `Campaign ${campaignId} does not support reminders for '${type}'`;
+    return helpers.sendResponse(res, 422, msg);
+  }
+
+  return app.locals.clients.northstar.Users.get('mobile', mobile)
+  .then((user) => {
+    mobilecommons.send_message(mobile, messageBody);
+    const msg = `Sent text for ${campaignId} ${type} to ${user.mobile}`;
+    logger.info(msg);
+    stathat('Sent relative text');
```
deadlybutter: Nitpick -- Not sure this should have the word relative in it

itsjoekent (approved):

deadlybutter:
Seems 👍

app/routes/campaigns.js
sergii-tkachenko: Ah, missed it, will fix it now. Thanks.

sergii-tkachenko:
I also renamed mobile to phone to be consistent: see signups.md and donorschoosebot.md.

app/models/Campaign.js
aaronschachter: @sergii-tkachenko and I chatted about naming these as:
````
messages:
  scheduled_relative_to_signup_date: <msg>
  scheduled_relative_to_reportback_date: <msg>
````

Comment (deadlybutter) [inline app/models/Campaign.js]: Oh gotcha -- sorry I thought the comment was in reply to that ^

File: app/routes/campaigns.js [hunk 7]
```diff
@@ -41,4 +46,51 @@ router.get('/:id', (req, res) => {
...
+  const messageBody = campaign.messages[type];
+  if (!messageBody) {
+    const msg = `Campaign ${campaignId} does not support reminders for '${type}'`;
+    return helpers.sendResponse(res, 422, msg);
+  }
+
+  return app.locals.clients.northstar.Users.get('mobile', phone)
+  .then((user) => {
+    mobilecommons.send_message(phone, messageBody);
+    const msg = `Sent text for ${campaignId} ${type} to ${user.mobile}`;
```
aaronschachter: We should avoid logging mobile numbers when possible. I'll be changing this to output the user.id in a future PR

sergii-tkachenko: Why?

Issue #775, opened 2017-01-27
Migrate content collections to Contentful 


aaronschachter:
Refs ideas in https://github.com/DoSomething/gambit/issues/764#issuecomment-275687754

* [x] Always GET campaign from Phoenix API to determine status, tagline, etc.
* [x] Create a `keyword` Contentful content type to deprecate `app.locals.keywords`
* [x] Create a `campaign` Contentful content type to deprecate `app.locals.campaigns`
* [x] Refactor `broadcasts` collection as a content type
* [x] Refactor `CampaignBot` as a Contentful campaign with title "default", or as a separate Configuration / Chatbot content type

Some work around deprecating `app.locals.campaigns`:
* [x] POST `signups` endpoint needs to query Phoenix to check if the given Campaign is closed
* [x] `helpers.isCampaignBotCampaign` should query Contentful to check if it has any published `production` keywords to determine if Campaign is available on Gambit 
* [ ] Group creation/update should happen on the fly? If we get a campaign ID and it doesn't have groups (or if they groups are not for the correct run) -- create groups and store on the Contentful Campaign

aaronschachter:
Some work around deprecating `app.locals.campaigns`:
* [ ] POST `signups` endpoint needs to query Phoenix to check if the given Campaign is closed, and query Contenful to check if it has any published `production` keywords
* [ ] GET `campaigns` endpoint logic should change to also query Contentful to check if it has any published `production` keywords to determine the `campaignbot` boolean property

aaronschachter:
`helpers.isCampaignBotCampaign` was removed here: https://github.com/DoSomething/gambit/commit/258c299919950af5d7513c59b210b74c7af2b9a5

For the last item, storing the Campaign Group ID's into Contentful -- I'm beginning to think each Campaigns request should just query out to the [Messaging Groups API](https://github.com/DoSomething/messaging-groups-api/) to get the latest ID's instead of maintaining the cached values in Contentful.

Issue #764, opened 2016-12-01
Investigate why campaign status wasn't updated


deadlybutter:
<img width="1042" alt="screen shot 2016-12-01 at 1 09 10 pm" src="https://cloud.githubusercontent.com/assets/897368/20806062/604caf50-b7c7-11e6-87e8-d3bbc7ddb40a.png">

aaronschachter:
This is likely an order of operations issue. If we're closing a Campaign and we remove the Campaign ID from `CAMPAIGNBOT_CAMPAIGNS` while its status is Active -- that Campaign status will never be updated to closed on [app start](https://github.com/DoSomething/gambit/blob/fbe1ead5067a50f137f1573cb5954e8efb212a0f/server.js#L138).

One idea to resolve is to change `loadCampaigns` to just update the cache for all Campaigns we have documents for.

@deadlybutter Can you update issue description with actual text? The screenshots of Slack rooms won't help surface this issue when searching.

aaronschachter:
I've pondering using `campaigns.findById` every time we need to load a Campaign instead of keeping them in memory on app start -- deprecating the `CAMPAIGNBOT_CAMPAIGNS` config variable altogether in favor of a new `campaignbot` boolean property on the Campaign model that admins can use to enable/disable CampaignBot campaigns. This would avoid needing to restart the app to turn campaigns on/off... although it would slow things down. We load keywords all in memory too -- which would be another Mongo query to execute if a `keyword` parameter is passed in an incoming chatbot request.

deadlybutter:
+1 to loading from DB instead of memory where possible

aaronschachter:
Closing this, as the investigation is done. This will be fixed when #775 is resolved.

PR #777, opened 2017-02-02
chatbot.fetchCampaign

aaronschachter:
#### What's this PR do?
* Adds a `fetchCampaign` helper function to `routes/chatbot` -- refactors to fetch a DS Campaign from the Phoenix API upon each request, instead of referencing the cached `app.locals.campaigns` Campaign document that only gets updated on app start

* Removes document properties like `rb_noun` and `rb_verb` that the CampaignBot no longer uses 
to render messages

* Clean up `objectID` variables as `objectId` (refs https://github.com/DoSomething/gambit/pull/697#discussion_r85598802)

* Adds TODOs for all code that references `app.locals.campaigns` -- which we'll likely get rid of always just `findById` (or better yet -- query Contentful)

Test the full flow for:
- Campaign that all CampaignBot default messages
- Campaign that has CampaignBot overrides in ithe `campaigns` DB

* By querying the Phoenix API for the campaign's status every request, we avoid gotchas like #764 -- where the Gambit needs to be restarted to refresh Campaign cache to update status

* Leaving out refactoring of the `campaign/:id/message` and /signups` endpoints for now to help keep these changes reviewable, and testable.

#### Relevant tickets
#775, #764

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -14,9 +14,14 @@ const logger = app.locals.logger;
  * Returns whether given Campaign id is enabled for CampaignBot.
  */
 module.exports.isCampaignBotCampaign = function (campaignId) {
-  const ids = process.env.CAMPAIGNBOT_CAMPAIGNS;
+  // TODO: We'll eventually determine if a CampaignBotCampaign is enabled if a keyword exists for
+  // its id (and its campaign.status is active).
+  const campaignDoc = app.locals.campaigns[campaignId];
+  if (!campaignDoc._id) {
+    return false;
+  }
 
-  return ids.includes(campaignId);
+  return true;
 };
 
 module.exports.isValidZip = function (zip) {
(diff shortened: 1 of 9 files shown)


lib/helpers.js
```diff
@@ -9,6 +9,21 @@ const logger = app.locals.logger;
...
  * Helper functions.
  */
+
+/**
+ * Returns whether given Campaign id is enabled for CampaignBot.
+ */
+module.exports.isCampaignBotCampaign = function (campaignId) {
+  // TODO: We'll eventually determine if a CampaignBotCampaign is enabled if a keyword exists for
+  // its id (and its campaign.status is active).
+  const campaignDoc = app.locals.campaigns[campaignId];
```
sergii-tkachenko:
```js
return !!app.locals.campaigns[campaignId]
```

See https://github.com/airbnb/javascript#coercion--booleans

sergiitk (approved):
Looks good, minor recommendation.

lib/helpers.js
aaronschachter: Wanted to check for the `_id` specifically to make sure whatever is stored in the array is a loaded document.. but maybe it's not necessary?

sergii-tkachenko: Then 
```
return !!app.locals.campaigns[campaignId].id
```

PR #781, opened 2017-02-03
Fix Campaigns endpoint

### LINKED ISSUE #780: GET campaigns returns empty set without campaignbot query param
👍 http://ds-mdata-responder-staging.herokuapp.com/v1/campaigns?campaignbot=true
👎 http://ds-mdata-responder-staging.herokuapp.com/v1/campaigns

Should also fix http://ds-mdata-responder-staging.herokuapp.com/v1/campaigns?campaignbot=false (we're only checking for existence of a `campaignbot` query param but not the value to filter by)

aaronschachter:
#### What's this PR do?
* Fixes #780, which was introduced in #777 in by `helpers.isCampaignBotCampaign`. When a Campaign is not a CampaignBot campaign, we'd get the error `Cannot read property '_id' of undefined`

* Calls `helpers.sendResponse` for errors to fix outputting errors

Upon staging deploy, verify: 
- http://ds-mdata-responder-staging.herokuapp.com/v1/campaigns 
- http://ds-mdata-responder-staging.herokuapp.com/v1/campaigns/362

Looking to change `isCampaignBotCampaign` logic by checking to see if there any keywords that exist for the given Campaign. If no keywords are found -- it's not on CampaignBot.

#### Relevant tickets
Fixes #780 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/campaigns.js
+++ b/app/routes/campaigns.js
@@ -54,7 +54,7 @@ router.get('/:id', (req, res) => {
 });
 
 /**
- * Sends SMS message to given phone number with the given Campaign and its message type .
+ * Sends SMS message to given phone number with the given Campaign and its message type.
  */
 router.post('/:id/message', (req, res) => {
   // Check required parameters.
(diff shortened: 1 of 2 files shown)


sergiitk (commented):
👍

PR #787, opened 2017-02-17
Deprecate app.locals.clients

aaronschachter:
#### What's this PR do?
* Adds `lib/phoenix` and `lib/northstar` and deprecates use of the `app.locals.clients`
* Deprecates wrapper `phoenix.fetchCampaign(id)` introduced in #786 in favor of directly calling Phoenix JS `Campaigns.get(id)`
* Cleans up nested `then.().catch()` in model functions

#### How should this be reviewed?
Full test of account creation, campaign signup, campaign reportback.

#### Any background context you want to provide?
Looking to additionally deprecate `app.locals.logger`, `app.locals.stathat`, `app.locals.db` in the same way. 

#### Relevant tickets
Refs @deadlybutter's suggestion in https://trello.com/c/H3nZ2UiA to deprecate usage of `app.locals`

#### Checklist
- [x] Tested on staging.

PR #789, opened 2017-02-17
Gambit Campaigns API fixes

### LINKED ISSUE #788: Add status back into GET campaigns response
The `status` property was removed from the `Campaign` model in #777 when we made the change to query the Phoenix API anytime we want information about a Campaign.

There are Quicksilver consumers that rely on its presence:
https://github.com/DoSomething/mbc-registration-mobile/pull/65
https://github.com/DoSomething/Quicksilver-PHP/pull/104

We'll need to query the Phoenix API for each campaign the Gambit GET `campaigns` endpoint returns. This will come in handy as well when it's time to built out functionality for [Slothbot](https://github.com/dosomething/slothbot) to support letting staff know what campaigns are live and their keywords.

aaronschachter:
#### What's this PR do?
* Adds `status` back into the Campaign response, via querying the Phoenix to get the campaign's status
* Fixes borked `campaign/:id/message` -- missed this in #787 

Verify **GET** `v1/campaigns` returns Campaign status
Verify **POST** `v1/campaigns/:id/message` returns expected results

Looking to use Contentful to get the copy for the relative reminder messages, instead of the `Campaign` model

#### Relevant tickets
Fixes #788

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/campaigns.js
+++ b/app/routes/campaigns.js
@@ -3,9 +3,12 @@
 const express = require('express');
 const router = express.Router(); // eslint-disable-line new-cap
 const Promise = require('bluebird');
+const UnprocessibleEntityError = require('../exceptions/UnprocessibleEntityError');
+
 const contentful = require('../../lib/contentful');
 const helpers = require('../../lib/helpers');
 const mobilecommons = rootRequire('lib/mobilecommons');
+const northstar = require('../../lib/northstar');
 const phoenix = require('../../lib/phoenix');
 const logger = app.locals.logger;
 const stathat = app.locals.stathat;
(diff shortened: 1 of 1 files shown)


sergiitk (approved):
👍

app/routes/campaigns.js
```diff
@@ -90,37 +100,62 @@ router.post('/:id/message', (req, res) => {
...
+      mobilecommons.send_message(phone, messageBody);
+      const msg = `Sent text for ${campaignId} ${type} to ${user.mobile}`;
+      logger.info(msg);
+      stathat('Sent campaign message');
+
+      return helpers.sendResponse(res, 200, msg);
+    })
+    // TODO: Catch UnprocessibleEntityError. Commenting this code out because it is
+    // throwing a "Class constructors cannot be invoked without 'new'"" error
+    // .catch(UnprocessibleEntityError, (err) => {
```
aaronschachter: This is the weirdest thing. I'm copying https://github.com/DoSomething/gambit/blob/3ca61cfc71be5c54e871ee271cfd67b7a1dfa8b1/app/routes/chatbot.js#L304 which doesn't throw this error.

sergii-tkachenko: Might be because something returns normal Promise, the one that isn't wrapped into Bluebird.

aaronschachter: Hmm good call... potentially the `phoenix` or `northstar` Superagent clients?

sergii-tkachenko: Might be. We need to look how filtered catch actually works in Bluebird: http://bluebirdjs.com/docs/api/catch.html#filtered-catch

PR #791, opened 2017-02-20
Source Relative Reminder Message copy from Contentful

### LINKED ISSUE #771: Handle POST reportback errors (Members stuck in 'Reason_Why' Loop)
_BUG_
-------------------
#### Current Behavior
- When some members (it's not affecting all users) respond to the last question in an SMS flow, they receive the same question instead of the report back confirmation message. We have seen this affect users on all of our active SMS campaigns - Take Back the Shelves, Senior Story Swap, Innovation Inspiration, and Don't Be a Sucker. During the same time period we have also seen successful SMS report backs on each of those campaigns. 

- Each staff member that was unable to receive the report back confirmation message (James, Lizzy, and Aki) had CGG as their source. Unclear if this is related to root cause, or just correlation. 

- Has affected 29 users since December 1. For context, there have been 369 SMS reportbacks from 12/21 to 12/22. 

#### Desired Behavior
- Members should have received the report back confirmation message when responding to 'Reason_why'. 

#### Steps to Replicate
- 

#### Why This Matters
- Members are unable to report back via SMS 

image

image

image

User affected by bug on the 19th - 

image

aaronschachter:
#### What's this PR do?
* Queries the Contentful Campaign content type to source Relative Reminder content from the `scheduledRelativeToSignupDateMessage` and `scheduledRelativeToReportbackDateMessage` fields, deprecating the `messages` property on the `Campaign` model.

    * If the Contentful Campaign for a given Campaign doesn't have values set for its `scheduledRelativeToSignupDateMessage` or `scheduledRelativeToReportbackDateMessage` -- use the copy defined on the Contentful Campaign with ID `default`.

    * Supports Liquid style variables for Campaign values, e.g. `{{title}}`, `{{tagline}}`

    * Removes the Northstar query from `campaigns/:id/message` -- we already have a phone number to send the message to and we aren't looking up any Signup information, so don't need to make a NS request (refs #766)

* Adds `helpers.replacePhoenixCampaignVars` function to DRY variable replacement in the `chatbot` and `campaigns` routes

    * Returns a `Promise` -- if we get a `{{keyword}}` variable, need to query Contentful to find the given Campaign's keyword

* Refactors `CampaignBotController` functions and `chatbot` route to return a message type as a string instead of the rendered message,  in prep for a future PR that will deprecate `CampaignBot` model in favor of sourcing message content from the Contentful Campaign

* Fixes #771 by returning an error in `CampaignBotController.postReportback` and checking for it in the `chatbot` route

Full test of all endpoints that deliver messages to end users: 📲  
* [x] Chatbot - from KEYWORD Signup to Reportback
* [x] POST Signups (External Signup)
* [x] POST Campaign Message

#### Relevant tickets
Fixes #771, refs #775 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -134,13 +134,6 @@ class CampaignBotController {
   }
 
   /**
-   * Wrapper function for logger.error(error)
-   */
-  error(req, err) {
-    logger.error(`${this.loggerPrefix(req)} ${err}:${err.stack}`);
-  }
-
-  /**
    * Returns whether current user has submitted a Reportback for the current campaign.
    * @param {object} req
    * @return {bool}
@@ -181,7 +174,12 @@ class CampaignBotController {
     return req.signup
       .postDraftReportbackSubmission()
       .then(() => 'menu_completed')
-      .catch(err => this.error(req, `postReportback ${err}`));
+      .catch((err) => {
+        const scope = err;
+        scope.message = `postReportback error:${err.message}`;
+
+        return scope;
+      });
   }
 
 }
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -4,7 +4,6 @@
  * Imports.
  */
 const logger = app.locals.logger;
-const campaignBot = app.locals.campaignBot;
 const helpers = rootRequire('lib/helpers');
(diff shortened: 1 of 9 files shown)


sergiitk (commented):
👍

PR #803, opened 2017-02-28
Check for keywords to determine if Campaign is running on Gambit

aaronschachter:
#### What's this PR do?
Instead of checking for the Campaign ID in `GAMBIT_CHATBOT_CAMPAIGNS` config variable, query Contentful to see if it has any keywords to determine if its enabled on Gambit.
* Removes deprecated `helpers.isCampaignBotCampaign` function
* removes deprecated `app.locals.campaigns` object

Post to Campaign Message and Signup endpoints with Campaigns that do/don't have keywords to verify expected behavior.

Next up is storing Messaging Group Id's in the Contentful Campaign content type instead of the `campaigns` collection, to deprecate the `Campaign` model entirely.

#### Relevant tickets
#775 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/models/Campaign.js
+++ b/app/models/Campaign.js
@@ -5,7 +5,6 @@
  */
 const mongoose = require('mongoose');
 const logger = app.locals.logger;
-const helpers = require('../../lib/helpers');
 const phoenix = require('../../lib/phoenix');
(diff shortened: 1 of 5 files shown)


sergiitk (approved):
👍

PR #804, opened 2017-02-28
Use mongoose.connect instead of mongoose.connection

aaronschachter:
#### What's this PR do?
* Refactor models to use the global Mongoose connection created when calling `mongoose.connect`

    * Gambit used to require two connections for a `config` database (stored content that admins would enter to associate Opt-in Path ID's with Campaigns) as well as the `mdata-responder` database we still use currently)

* Deprecate `app.locals.db` object, instead just `require` each Model file needed

* Removes loading DonorsChooseBot for now -- will open a separate PR to load DonorsChooseBot content from a JSON file (doesn't seem worth the lift of migrating to Contentful, as it's uncertain if we'll ever run it again) 

Test all functionality on the `chatbot`, `campaigns`, and `signups` endpoints.

Eventually going to get rid of all `app.locals` properties.

#### Checklist
- [ ] Tested on staging.


sergiitk (commented):
👍

Issue #811, opened 2017-03-10
Gambit is crashing when New Relic is enabled


aaronschachter:
Gambit keeps crashing with `at=error code=H10 desc="App crashed"`

Let's try turning off New Relic. Here's a stack trace that leads us to think it's worth a shot:
````
2017-03-10T15:07:12.115644+00:00 app[web.1]:        function: 'Connection.connect',
2017-03-10T15:07:12.115645+00:00 app[web.1]:        line: 389,
2017-03-10T15:07:12.115645+00:00 app[web.1]:        method: 'connect',
2017-03-10T15:07:12.115646+00:00 app[web.1]:        native: false },
2017-03-10T15:07:12.115646+00:00 app[web.1]:      { column: 14,
2017-03-10T15:07:12.115647+00:00 app[web.1]:        file: '/app/node_modules/mongodb-core/lib/connection/pool.js',
2017-03-10T15:07:12.115647+00:00 app[web.1]:        function: '_createConnection',
2017-03-10T15:07:12.115639+00:00 app[web.1]:        native: false },
2017-03-10T15:07:12.115648+00:00 app[web.1]:        line: 1059,
2017-03-10T15:07:12.115648+00:00 app[web.1]:        method: null,
2017-03-10T15:07:12.115649+00:00 app[web.1]:        native: false },
2017-03-10T15:07:12.115650+00:00 app[web.1]:      { column: 13,
2017-03-10T15:07:12.115650+00:00 app[web.1]:        file: '/app/node_modules/mongodb-core/lib/connection/pool.js',
2017-03-10T15:07:12.115651+00:00 app[web.1]:        function: null,
2017-03-10T15:07:12.115651+00:00 app[web.1]:        line: 1151,
2017-03-10T15:07:12.115652+00:00 app[web.1]:        method: null,
2017-03-10T15:07:12.115652+00:00 app[web.1]:        native: false },
2017-03-10T15:07:12.115653+00:00 app[web.1]:      { column: 39,
2017-03-10T15:07:12.115653+00:00 app[web.1]:        file: '/app/node_modules/mongodb-core/lib/connection/pool.js',
2017-03-10T15:07:12.115654+00:00 app[web.1]:        function: 'waitForAuth',
2017-03-10T15:07:12.115654+00:00 app[web.1]:        line: 1082,
2017-03-10T15:07:12.115655+00:00 app[web.1]:        method: null,
2017-03-10T15:07:12.115655+00:00 app[web.1]:        native: false },
[...]

PR #813, opened 2017-03-10
Automatically update New Relic

aaronschachter:
#### What's this PR do?
Update New Relic to 1.37.1 which might help #811


sergii-tkachenko:
I just double checked, this PR actually will do nothing.
Initial `^1.28.3` does already include latest v `1.37.2`.

PR #814, opened 2017-03-10
Pin New Relic version to 1.37.0 

aaronschachter:
#### What's this PR do?
Pins version to 1.37.0 to try re-enabling New Relic, and avoiding #811


sergiitk (commented):
👍

PR #825, opened 2017-03-22
Add function names to error messages

### LINKED ISSUE #816: Add service name to network error messages
When an external network request fails, we log the error message -- but it's tough to know which request of the many we make is throwing the error. Examples:

* `{"error":{"code":500,"message":"Bad Gateway"}}`
* `Connection failed. Errno::ETIMEDOUT CC: 3. E: ''`
* `{"error":{"code":500,"message":"Internal Server Error"}}`

I've started adding prefixes into some functions to indicate where the error is occurring, e.g:\
* [Contentful requests](https://github.com/DoSomething/gambit/blob/488fac0dbe70e7ec2a85d6cff5b9ea7edf0d6eb3/lib/contentful.js#L50)
* [posting Reportbacks to Phoenix](https://github.com/DoSomething/gambit/blob/488fac0dbe70e7ec2a85d6cff5b9ea7edf0d6eb3/app/controllers/CampaignBotController.js#L179) 

This issue is to update all functions to better indicate which request is throwing the error returned.

aaronschachter:
#### What's this PR do?
Adds a suffix to error messages in our Model static functions (and a new Phoenix `fetchCampaign` helper function) to help troubleshoot where network errors come from.

Try to break things. One way to see the prefix in action is by attempting to post a signup keyword (for a campaign you haven't signed up for), but with an invalid `DS_PHOENIX_API_PASSWORD` config variable. Instead of a 500 error, the returned response should look like:
````
{
 "error": {
   "code": 401,
   "message": "Signup.post error:Unauthorized"
 }
}
````

We mostly see these errors when during high volume after sending out a broadcasts.

#### Relevant tickets
Fixes #816 , Refs #810 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/models/Signup.js
+++ b/app/models/Signup.js
@@ -89,7 +89,9 @@ signupSchema.statics.lookupById = function (id) {
           return reject(notFoundError);
         }
 
-        return reject(err);
+        const scope = err;
+        scope.message = `Signup.lookupById error:${err.message}`;
+        return reject(scope);
       });
   });
 };
@@ -129,8 +131,10 @@ signupSchema.statics.lookupCurrent = function (user, campaign) {
       .then(signupDoc => resolve(signupDoc))
       .catch((err) => {
         app.locals.stathatError(statName, err);
+        const scope = err;
+        scope.message = `Signup.lookupCurrent error:${err.message}`;
 
-        return reject(err);
+        return reject(scope);
       });
   });
 };
@@ -171,8 +175,10 @@ signupSchema.statics.post = function (user, campaign, keyword) {
       .then(signupDoc => resolve(signupDoc))
       .catch((err) => {
         app.locals.stathatError(statName, err);
+        const scope = err;
+        scope.message = `Signup.post error:${err.message}`;
 
-        return reject(err);
+        return reject(scope);
       });
   });
 };
@@ -203,7 +209,12 @@ signupSchema.methods.createDraftReportbackSubmission = function () {
 
         return resolve(updatedSignup);
       })
-      .catch(err => reject(err));
+      .catch((err) => {
+        const scope = err;
+        scope.message = `Signup.post error:${err.message}`;
+
+        return reject(scope);
+      });
   });
 };
 
@@ -257,7 +268,10 @@ signupSchema.methods.postDraftReportbackSubmission = function () {
         submission.failed_at = dateSubmitted;
         submission.save();
 
-        return reject(err);
+        const scope = err;
+        scope.message = `Signup.post error:${err.message}`;
+
+        return reject(scope);
       });
   });
 };
--- a/app/models/Campaign.js
+++ b/app/models/Campaign.js
@@ -50,7 +50,12 @@ campaignSchema.statics.lookupByIds = function (campaignIds) {
 
         return resolve(Promise.all(promises));
       })
-      .catch(error => reject(error));
+      .catch((err) => {
+        const scope = err;
+        scope.message = `Campaign.lookupByIds error:${err.message}`;
+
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: app/models/Signup.js [hunk 1]
```diff
@@ -129,8 +131,10 @@ signupSchema.statics.lookupCurrent = function (user, campaign) {
       .then(signupDoc => resolve(signupDoc))
       .catch((err) => {
         app.locals.stathatError(statName, err);
```
rapala61: should we send this new custom error message to stathat too? It could probably help troubleshoot in the future.

aaronschachter: We're currently sending the status code for external network requests through `stathatError`, which help provide some insight into some of those errors we get back after broadcasts. 

@rapala61 I'm not opposed to adding the error message in the stat name as well.. but wonder if it might be redundant (as the code should always match). It could potentially catch some non-network errors like Mongo writes, but New Relic would be logging those. 

screen shot 2017-03-22 at 8 35 58 am

rapala61: Ok. I was looking at this tab in stathat: https://www.stathat.com/v/stats/Zxka and thought the message would help clear up what this error means "bad gateway". I maybe just have to get more familiar with how all this errors are being reported.

aaronschachter: I was looking at the `chatbot` router to refresh my memory on how that Bad Gateway message is getting populated -- we're logging error messages to Stathat [here](https://github.com/DoSomething/gambit/blob/d427963b9c419c6ac571fe4b6d01460c7fb8bc22/app/routes/chatbot.js#L368), so the newly prefixed message should hopefully show up `WhateverIsCallingMe: Bad Gateway`.

rapala61: Cool, thanks!

rapala61 (approved):

aaronschachter:
Our memory usage spiked after deploying this (v290)
screen shot 2017-03-23 at 9 56 58 am

I've rolled back the Heroku deployment and memory immediately down to 8%, will keep 👀  on it
screen shot 2017-03-23 at 9 23 14 am

aaronschachter:
22 hours later and it seems safe to say the issue is somewhere in this PR 🤕 
screen shot 2017-03-24 at 7 37 37 am

rapala61:
yes... but for some obscure reason. I'm going to attack this all day today. I'll update you later on what I find. That code would mostly impact users that are reporting back. That is a little cumbersome to test at scale, but I'll try to devise some strategy to do that.

aaronschachter:
Thanks @rapala61 -- really thankful you can take a look, let me know if I can be helpful. There are 2 main changes in this PR:

1. creating copies of a caught error,  and altering the `message`

2. replacing all calls to `phoenix.Client.campaigns.get` with the helper function `phoenix.fetchCampaign`, for the sake the first change (customizing the caught error message)

rapala61:
Got it. Could you also rollback the release in Github? Thanks!

aaronschachter:
Chatted with @rapala61 about ☝️ in Slack -- holding off on reverting these commits for now, as it'd be a bit messy -- and we're still unsure of the cause. 

I was noticing that the memory usage jumped up around 3 hours after deployment, which seems to be timed with when Quicksilver should have [sent Relative Reminder messages](https://github.com/dosomething/gambit/wiki/quicksilver#scheduled-relative-messages) out. We've redeployed 3.6.1 (v292) to see if we see the same spike around 12 PM EST today. The past half-hour it's been running on production hasn't shown a increase in mem usage... but we'll 👁 

screen shot 2017-03-24 at 9 02 11 am

Issue #744, opened 2016-11-12
Deprecate CampaignBotController


aaronschachter:
Someone new to the project might wonder why the heck the Reportback Start and Continue functions live in a `CampaignBotController` class, and not just in the `route/chatbot`.

It's because I haven't gotten around to refactoring it to move it into the `route/chatbot`, and to deprecate the `CampaignBotController` class entirely -- like we did for DonorsChooseBot in #720, #722 

The CampaignBotController used to be a gigantic one-function-calls-another-function legacy style class -- #632 started on this based on @deadlybutter 's advice in https://github.com/DoSomething/gambit/pull/631#discussion_r80523908. Other related PR's toward deprecating `CampaignBotController` include #686, #723, and likely some others.

This isn't super high priority, but it'd be really nice to get done at some point to make it painfully obvious where all code shoudl go (either the `route` or the `model`, but no more `controllers`)

* [ ] Move validation/save into `ReportbackSubmission` model
* [ ] Move start/continue logic into `routes/chatbot`

PR #826, opened 2017-03-22
Deprecate CampaignBotController.createReportbackSubmission

aaronschachter:
#### What's this PR do?
Starts chipping away at #744 by removing the extraneous `CampaignBotController.createReportbackSubmission` call in the `chatbot` router, directly calling `Signup.createDraftReportbackSubmission` instead.

* Remove your Signup for a given Campaign
    * Use the Phoenix Unsignup form to delete from Phoenix DB
    * Delete your local `signups` document for your user/campaign if exists
* Post the keyword for the Campaign
* Verify Signup is created
* Post the Reportback command
* Verify the Signup's `draft_reportback_submission` is populated, and the Ask Quantity message is returned

Opening this up as a bite-sized PR that's easy to review before starting to take on moving the CampaignBotController `collectReportbackProperty` and `continueReportbackSubmission` functions into the `ReportbackSubmission` model and `chatbot` router.

#723 -- old PR moving many more `CampaignBotController` functions out into model classes

#### Relevant tickets
#744 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/controllers/CampaignBotController.js
+++ b/app/controllers/CampaignBotController.js
@@ -109,24 +109,6 @@ class CampaignBotController {
   }
 
   /**
-   * TODO: Move into router.
-   * Creates new ReportbackSubmission model and updates Signup model's draft.
-   * @param {object} req - Express request
-   * @return {string} - Message to send and begin collecting Reportback data.
-   */
-  createReportbackSubmission(req) {
-    this.debug(req, 'createReportbackSubmission');
-
-    return req.signup
-      .createDraftReportbackSubmission()
-      .then(() => {
-        this.debug(req, `updated signup:${req.signup._id.toString()}`);
-
-        return this.collectReportbackProperty(req, 'quantity', true);
-      });
-  }
-
-  /**
    * Wrapper function for logger.debug(msg)
    */
   debug(req, msg) {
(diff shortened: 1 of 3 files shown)


aaronschachter:
@rapala61 was testing this branch and found that a `reportback_submission` document was being created with a null user. I've been looking into this, it seems to happen I create a new user and send a signup keyword. and it's happening on `develop`.

screen shot 2017-03-22 at 2 17 42 pm
Checking into prod db and I see 6 `reportback_submission` documents with a `null` user, all from March 2017.
`db.getCollection('reportback_submissions').find( { user : { $type: 10 } } )`

In theory, this shouldn't affect anything except being able to query `reportback_submissions` by user for our own investigation/reporting purposes. The chatbot router will load a Signup for a user, then find the reportback submission stored on the `signup.draft_reportback_submission.`

What's odd is there are 47 Signups with null `user`  -- they have Campaign id's of campaigns which are currently closed, but we don't have the timestamp stored in mongo (as the Pheonix signup id would store the canonical time created). Will need to start looking into these id's to investigate what the cause is. 
`db.getCollection('signups').find( { user : { $type: 10 } } )`

screen shot 2017-03-22 at 2 22 33 pm

PR #829, opened 2017-03-24
Add New Relic Native Metrics

aaronschachter:
#### What's this PR do?
Installs the [New Relic Native Metrics](https://docs.newrelic.com/docs/agents/nodejs-agent/supported-features/node-vm-measurements) module to "gain insights into garbage collection and event loop behavior", turning this:

screen shot 2017-03-24 at 9 35 51 am

into this:

screen shot 2017-03-24 at 9 31 27 am

* Check it out on staging upon deploying this branch, or enable New Relic locally and set it to the `Gambit Local` app, which is where screenshot above is from.

```
NEW_RELIC_ENABLED=true 
NEW_RELIC_APP_NAME=Gambit Local
```

#### Relevant tickets
https://github.com/DoSomething/gambit/pull/825#issuecomment-288776394

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   "dependencies": {
     "@dosomething/northstar-js": "git://github.com/DoSomething/northstar-js.git",
     "@dosomething/phoenix-js": "git://github.com/DoSomething/phoenix-js.git",
+    "@newrelic/native-metrics": "^2.1.0",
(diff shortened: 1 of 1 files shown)


rapala61 (approved):
Nice, this is exactly what I'm doing, but this is better.

Issue #827, opened 2017-03-23
Some signup and reportback_submission documents have a null user


aaronschachter:
_BUG_
-------------------
#### Current Behavior
Refs https://github.com/DoSomething/gambit/pull/826#issuecomment-288544291

* A small number of `signup` documents (48) have a null `user` in production.
    * db.getCollection('signups').find( { user : { $type: 10 } } )
* When testing signups with a new user (and fake mobile number), the draft `reportback_submission` is stored with a null user.
    * There are only 6 `reportback_submissions` documents with a null user on production

#### Desired Behavior
The `user` property should store the Signup or Reportback Submission User's Northstar ID.

#### Steps to Replicate
Don't know yet

#### Why This Matters
For Signups with a null user -- I'm not entirely sure it does, but want to document this somewhere. I think this happens when Quicksilver posts external Signups to the Gambit `/signups` endpoint, and the GET request to find the Northstar user fails (e.g. https://dosomething.slack.com/archives/C1V0M6RPE/p1490230443751810). I've been digging into Signups and Users here: https://gist.github.com/aaronschachter/9725c8d885808361f6751d40ae39e473 and so far a common thread seems to be they are external signups.

For Reportback Submissions with a null user -- let's pinpoint why this is happening.

aaronschachter:
Small number of `signup:0` that show up in the logs, which occur when can replicate this bug.
```
» 21 Mar 2017 12:18:16.229 125 <190>1 2017-03-21T19:18:16.181248+00:00 app web.1 - - [32minfo[39m: loaded signup:0 Context
» 21 Mar 2017 12:49:29.823 138 <190>1 2017-03-21T19:49:29.578156+00:00 app web.1 - - [32minfo[39m: Signup.post created signup:0 Context
» 21 Mar 2017 12:49:29.823 125 <190>1 2017-03-21T19:49:29.582152+00:00 app web.1 - - [32minfo[39m: loaded signup:0 Context
» 21 Mar 2017 12:55:17.566 138 <190>1 2017-03-21T19:55:17.494226+00:00 app web.1 - - [32minfo[39m: Signup.post created signup:0 Context
» 21 Mar 2017 12:55:17.566 125 <190>1 2017-03-21T19:55:17.504371+00:00 app web.1 - - [32minfo[39m: loaded signup:0 Context
[...]

PR #830, opened 2017-03-27
Query Phoenix instead of Northstar for Signups

aaronschachter:
#### What's this PR do?
Updates `Signup` model per https://github.com/DoSomething/phoenix-js/pull/19 to work toward a fix for #827. Currently we [get a null `user.id` back in the Signups response](https://github.com/dosomething/gambit/issues/827#issuecomment-289496978), so #827 isn't fixed until that's resolved.

Post a keyword to chatbot to complete a campaign: 
* Your Signup should load correctly and you should be able to continue the conversation, but... 
* Posting your Reportback should fail with `postReportback error:Cannot read property 'phoenix_id' of null`.
    * Gambit tries to load a document for a null user, and doesn't have the User properties it needs (like the `phoenix_id`)

#### Relevant tickets
* #827
* https://github.com/DoSomething/phoenix-js/pull/19
* https://github.com/DoSomething/phoenix/issues/7329
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/models/Signup.js
+++ b/app/models/Signup.js
@@ -42,19 +42,19 @@ const signupSchema = new mongoose.Schema({
 
 });
 
-function parseNorthstarSignup(northstarSignup) {
+function parsePhoenixSignup(phoenixSignup) {
   const data = {
-    _id: Number(northstarSignup.id),
-    user: northstarSignup.user,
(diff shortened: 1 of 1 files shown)


sergiitk (approved):
👍

Issue #810, opened 2017-03-09
Gambit timeouts during SMS Broadcasts and Niche Import


jamjensen:
#### Description
- We've recently seen a spike in request timeouts, which most often happens after we send SMS broadcasts or when we import niche users.  The app crashes and the user receives the mdata error response message.  

- Heroku outlines potential causes of request timeouts here: https://devcenter.heroku.com/articles/request-timeout#timeout-behavior 

- Potential solutions include background jobs, which Heroku outlines here: https://devcenter.heroku.com/articles/background-jobs-queueing

Refs #704

jamjensen:
![](https://github.trello.services/images/mini-trello-icon.png) [GAMBIT: Investigate solutions for timeout requests](https://trello.com/c/nATpa5UU/293-gambit-investigate-solutions-for-timeout-requests)

jamjensen:
![](https://github.trello.services/images/mini-trello-icon.png) [GAMBIT: Investigate solutions for timeout requests](https://trello.com/c/nATpa5UU/293-gambit-investigate-solutions-for-timeout-requests)

jamjensen:
*Notes from the Gambit Request Timeout Meeting*

AS: "Hunch is that the culprit = external calls to Phoenix / Northstar"
MH: "New Relic would show us how calls are affecting gambit performance"

When understanding the issue, biggest things to look for are:
- What slows down NS, Phoenix and the app 

Examples of users with calls taking >30 seconds:
1. 95% of transaction cost is talking to contentful... if we're hitting stuff hard on contentful we should look into why a call to contentful would take that long.... AS: "can be two calls to contentful bc the keywords and campaigns live in separate documents... if KW comes in during incoming request then when we render the message we query contentful to see if it has a field set in campaign flow, if not then query again for the default campaign." MH: "possibly that these are one offs." 

2. Post to the signups endpoint: MH "really long request to NS for user ID... signing them up for a campaign or pulling their info (i.e. phone number)... why is it taking over 30 seconds?

- RP: "My theory is that the endpoint itself doesn't become unstable but rather the application. Would be good to know if we're stressing the system in a different way... e.g. If I generate as many requests as my computer can handle, but it wasn't because MoCo could send them to me but rather than my computer couldn't handle it.. so it's possible that we're stressing our only dyno." 

*Takeaways:*
1. Slow requests to contentful (why would a request to your super fast CDN take so long?)
2. NS gets bogged down

*Next Steps:*
[...]

PR #836, opened 2017-03-30
WIP: Respond immediately with success for chatbot requests

### LINKED ISSUE #776: Refactor Gambit chatbot endpoint to return success upon message received
Refs https://github.com/DoSomething/gambit/issues/704#issuecomment-275727492

If we received a message, send a 200 "Message received" message back to Mobile Commons, to avoid timeout requests due to slow Northstar / Phoenix queries. We'll need to rely on New Relic and Stathat events/alerts to look out for events like failed Signups / user registrations (vs the mData alert).

aaronschachter:
#### What's this PR do?
This is a work in progress / proof of concept to experiment with the suggestion in #776: sending a 200 response immediately for an incoming `chatbot` request, in an attempt to avoid H12 timeouts https://github.com/dosomething/gambit/issues/810#issuecomment-290155928

The incoming response is logged immediately in the `bot_requests` collection and saved to a `currentBotRequest` constant, whose ID is returned in our success response:

````
{
  "success": {
    "code": 200,
    "message": "Queued request:58dd13acd589fd7377624257"
  }
}
````

The chatbot router `post` callback will still continue to process the incoming request after sending the success response back, eventually sending back a reply message via `user.postMobileCommonsProfileUpdate` or `mobilecommons.send_message`, and update the `currentBotRequest` accordingly:

````
{
    "_id" : ObjectId("58dd13acd589fd7377624257"),
    "client" : "mobilecommons",
    "body" : {
        "phone" : "[removed]",
        "mms_image_url" : "http://www.fox.com/sites/default/files/X-Files_hero.png",
        "args" : "p"
    },
    "bot_type" : "chatbot",
    "created_at" : ISODate("2017-03-30T14:17:16.975Z"),
    "__v" : 0,
    "bot_response_message" : "@dev:  Sweet! First, what's the total number of Books you Collected?\n\nSend the exact number back.",
    "bot_response_type" : "ask_quantity",
    "campaign_id" : 2299,
    "user_id" : "5547be89469c64ec7d8b518d"
}
````
##### Errors

When an error occurs - it's on Gambit to inform the end user.  I've been wondering whether we really need to do that (or just don't respond). Current behavior is the Gambit API sends an error response back, which triggers sending the mData error message defined in Mobile Commons

##### New Relic

This is where I'm hesitant to merge and deploy this right away (after a full integration test). From what I can see in the Gambit Local New Relic app with my local testing -- I'm not seeing any of the external requests to Phoenix, which I'm assuming (but not 100% positive) is because Gambit has already sent back the success response. This feels like the big tradeoff of going this route, and hits home for the "real" solution here is a background / Message Broker job.

screen shot 2017-03-30 at 7 31 19 am

screen shot 2017-03-30 at 7 31 41 am

Preliminary proof of concept review:
* Does this make sense? Does the code look ok?
* Test locally or on staging and inspect New Relic 

If we're into it and want to merge into `develop`, we'll need to post to the `chatbot` endpoint for all possible use cases to ensure it's not broken. I haven't done this yet as this branch is still in proof-of-concept phase.

We'll need a similar approach for responding once we have background tasks in place, so this seemed like an idea worth exploring.

#### Relevant tickets
Fixes #776 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -79,9 +79,10 @@ router.post('/', (req, res) => {
   BotRequest.log(req, 'chatbot')
     .then((botRequestDoc) => {
       currentBotRequest = botRequestDoc;
-      logger.info(JSON.stringify(currentBotRequest.body));
+      const id = botRequestDoc._id.toString();
+      logger.info(`created botRequest:${id} data:${JSON.stringify(currentBotRequest.body)}`);
 
-      return helpers.sendResponse(res, 200, `Message ${botRequestDoc._id.toString()} queued.`);
+      return helpers.sendResponse(res, 200, `Queued request:${id}`);
     })
     .catch(err => helpers.sendResponse(res, 500, err.message));
 
@@ -117,6 +118,7 @@ router.post('/', (req, res) => {
         logger.info(`loaded user:${user._id}`);
         scope.user = user;
         newrelic.addCustomParameters({ userId: user._id });
+        currentBotRequest.user_id = scope.user._id;
 
         if (scope.broadcast_id) {
           return contentful.fetchBroadcast(scope.broadcast_id)
@@ -207,6 +209,7 @@ router.post('/', (req, res) => {
     .then((campaign) => {
       scope.campaign = campaign;
       newrelic.addCustomParameters({ campaignId: campaign.id });
+      currentBotRequest.campaign_id = scope.campaign.id;
 
       if (phoenix.isClosedCampaign(campaign)) {
         throw new ClosedCampaignError(campaign);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/routes/chatbot.js [hunk 1]
```diff
@@ -86,33 +75,20 @@ router.post('/', (req, res) => {
...
-      const msg = `undefined process.env.${configVar}`;
-      stathat(`error: ${msg}`);
-      logger.error(msg);
-      configured = false;
-    }
-  });
-
-  if (!configured) {
-    return res.sendStatus(500);
-  }
```
aaronschachter: This setup check should move into somewhere like `server.js`.

app/routes/chatbot.js
```diff
@@ -86,33 +75,20 @@ router.post('/', (req, res) => {
...
-
-  if (!req.body.phone) {
-    stathat('error: undefined req.body.phone');
```
aaronschachter: TODO:  Create a separate middleware function to valid expected body parameters. Felt relatively safe to remove for now since all incoming requests to `chatbot` are from Mobile Commons mData.

aaronschachter:
We'll likely want the same functionality for the[ endpoints that Quicksilver posts to](https://github.com/DoSomething/gambit/wiki/Quicksilver) as well.

app/routes/chatbot.js
rapala61: Saving to BotRequest would also fit nicely as its own middleware before it reaches the route.

app/routes/chatbot.js
```diff
@@ -86,33 +75,20 @@ router.post('/', (req, res) => {
...
+  let currentBotRequest;
+  BotRequest.log(req, 'chatbot')
```
rapala61: Saving to BotRequest would also fit nicely as its own middleware before it reaches the route.

rapala61: currentBotRequest could be injected in the req object and then passed to the route.

app/routes/chatbot.js
rapala61: would failing this check bring the whole app down? If not, it could be moved to its own middleware that would affect only a determined route. Otherwise I agree it should go somewhere in `server.js`.

aaronschachter:
@rapala61 and I discussed these changes in Slack, and after a day or two of H12-free broadcasting to the 30k list. I've generally been restarting dynos before the AM broadcasts -- we're not certain if it works (investigating in #837, #810) but it's a hack that seems to be a better choice than these drastic changes which drop a lot of the valuable metrics we collect from New Relic about response times.

Issue #714, opened 2016-11-02
Instrument basic StatHat coverage


mshmsh5000:
Let's get StatHat coverage for the basic events in Gambit: signups, reportbacks, maybe errors?

aaronschachter:
Keeping this open until we add alerts for any misconfiguration that can go down (and trigger emails). Some examples:
- Keyword is live for a closed (or unpublished) Campaign
- Keyword doesn't have a matching Campaign for it in the active CampaignBot campaigns
- The Signup Broadcast Campaign is closed or not found

jamjensen:
I wanted to have more of a dashboard feel when trying to figure out what's happening in Gambit so I made some StatHat cards. 

- All error messages sent at different stages of user flow: https://www.stathat.com/cards/ZnGN4DygwRmz 
- Number of completed report backs vs number of report backs started: https://www.stathat.com/cards/FoOK5awF6lx1
- Signup Broadcast performance: https://www.stathat.com/cards/5ud5uYFBcAiV 
- Total number of signups to campaignbot: https://www.stathat.com/cards/JEfBaT43Rr4V
- User flow broken down by each message (Quantity, Photo, Caption, Why Particpated): https://www.stathat.com/cards/0oH9GyyId8KS

mshmsh5000:
Beautiful!


CTO
@mshmsh5000 <http://twitter.com/mshmsh5000> @dosomething
<http://twitter.com/dosomething>

aaronschachter:
I think we're in decent shape to close this -- we're tracking errors in Stathat, New Relic -- and errors returned to our mData show up in Slack, with more descriptive messages per #825.

PR #840, opened 2017-04-03
V8 profiler

### LINKED ISSUE #837: Collect Heap data to analyze and find the reason for Gambit Memory Leak
_BUG_
-------------------
#### Current Behavior
- There is an steady increase in memory over time in Gambit.

#### Desired Behavior
- Collect Heap Snapshots from Gambit live and upload to S3 for offline analysis.

#### Steps to Replicate
- It happens all the time just by being up and running in Staging and Production. 
- It seems to have started after this pull request: https://github.com/DoSomething/gambit/pull/825

#### Why This Matters
- It might have a connection to some timeouts that are not due to external requests. See: https://github.com/DoSomething/gambit/issues/810#issuecomment-290148244

- See:
   - https://github.com/DoSomething/gambit/issues/810#issuecomment-290148244
   - https://github.com/DoSomething/gambit/issues/810#issuecomment-290271825

_FEATURE OVERVIEW_
-------------------
#### User Story

As a O'doyle dev, I want to investigate Gambit memory leak so that Gambit doesn't crash.

Heroku Dyno restarts before broadcasts and schedule restarts.

rapala61:
#### What's this PR do?
It adds the ability to take Heap Snapshots of the app at determined intervals. 

- Install Gambit locally
- checkout this branch `v8-profiler`
- Add the following env variables:
    - AWS_ACCESS_KEY_ID=**Look in lastpass Gambit S3 secured document**
    - AWS_SECRET_ACCESS_KEY=**Look in lastpass Gambit S3 secured document**
    - V8_PROFILER_NAME_PREFIX=local.v8.heap.snapshot
    - V8_PROFILER_ENABLED=true
    - V8_PROFILER_DEFAULT_INTERVAL_MS=20000. **This will create a snapshot every 20 seconds.**
- A log message should confirm the upload was successful:
    - `10:28:38 AM web.1 |  debug: Snapshot upload successful. File name: v8.heap.snapshot_worker1_1491316082756.heapsnapshot`
- If you have s3 access: Confirm that you see a new snapshot created at: [S3 v8-profiler Bucket](https://console.aws.amazon.com/s3/buckets/v8-profiler.dosomething.org/?region=us-east-1&tab=overview)
We have a memory leak in Gambit that has not been able to be reproduced locally. After trying a lot of unsuccessful ways to reproduce locally, input from the #devops team was required. @pris insights were to invest some time and build the capability of taking Heap Snapshots into Gambit so we could analyze the Heap Snapshot offline. As an added bonus, we can keep this feature handy and turn it off/on as we need via env variables.

#### Relevant tickets
Fixes #837 

#### Checklist
- [ ] Tested on staging.


lib/tools/profiler.js
```diff
@@ -0,0 +1,113 @@
+'use strict';
+
+const v8Profiler = require('v8-profiler');
+const EventEmitter = require('events');
+
+// 30 min
+const DEFAULT_INTERVAL_MS = 1800000;
+const NAME_PREFIX = 'v8.heap.snapshot';
```
aaronschachter: Does it make sense to conditionally check for config variables here, to avoid code changes if we want to tweak this?

```
const DEFAULT_INTERVAL_MS = process.env.PROFILER_DEFAULT_INTERVAL_MS || 1800000;
const NAME_PREFIX = process.env.PROFILER_NAME_PREFIX || 'v8.heap.snapshot';
```

lib/tools/profiler.js
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * It generates the name of the snapshot.
+   * The snapshot's name follows the next pattern:
+   *    <PREFIX>_<TAG>_<EPOCH TIMESTAMP>
+   *    Example: v8.heap.snapshot_iamatag_1491242331638
+   *
+   * @return {String}
+   */
+  _getName() {
```
aaronschachter: Minor tweak, but if we passed the timestamp as a parameter to `getName`, we could write a unit test for this function.

server.js
```diff
@@ -59,6 +59,55 @@ function startWorker(id) {
...
+
+  if (process.env.V8_PROFILER_ENABLED &&
+      process.env.AWS_ACCESS_KEY_ID &&
+      process.env.AWS_SECRET_ACCESS_KEY) {
+    logger.info('HEAP SNAPSHOT PROFILER IS ENABLED.');
+
+    const profiler = new Profiler(null, `worker #${id}`, 'heapsnapshot');
+    const uploader = new Uploader({
+      params: {
+        Bucket: 'v8-profiler.dosomething.org',
```
aaronschachter: Same here, something like: `Bucket: process.env.UPLOADER_BUCKET || 'v8-profiler.dosomething.org'` would allow tweaks without code changes...

.env.example
```diff
@@ -3,6 +3,9 @@ LOGGING_LEVEL=info
 
 WEB_CONCURRENCY=1
 
+AWS_ACCESS_KEY_ID=keygoeshere
+AWS_SECRET_ACCESS_KEY=keygoeshere
```
aaronschachter: Worth adding `V8_PROFILER_ENABLED=false` ?

aaronschachter: Hm - Actually I think that check [would still run the profiler,](https://github.com/DoSomething/gambit/blob/8dc132d0e0999f9cb4d85a10d618cf2d33f631eb/server.js#L69) as the config variable would be set to a string. Just seems like we'd want to add `V8_PROFILER_ENABLED` if we're adding the `AWS` variables.

server.js
```diff
@@ -59,6 +59,55 @@ function startWorker(id) {
...
+
+      /**
+       * This is the multipart stream that will upload the snapshot to S3
+       */
+      uploadStream
+        .on('error', (error) => {
+          logger.info(`Upload Stream error: ${error}`);
+          snapshot.delete();
+        })
+        .on('uploaded', () => snapshot.delete());
```
aaronschachter: It'd be useful to get a log message upon upload success, maybe just `logger.debug`. I don't have access to the Gambit S3 yet, so can't verify whether anything's been uploaded by looking at the logs. Not sure if it makes sense to `debug` here or in `snapshot.export().pipe` (or both)...

server.js
```diff
@@ -59,6 +59,55 @@ function startWorker(id) {
...
+  /**
+   * Profiler setup
+   */
+
+  if (process.env.V8_PROFILER_ENABLED &&
+      process.env.AWS_ACCESS_KEY_ID &&
+      process.env.AWS_SECRET_ACCESS_KEY) {
+    logger.info('HEAP SNAPSHOT PROFILER IS ENABLED.');
+
+    const profiler = new Profiler(null, `worker #${id}`, 'heapsnapshot');
```
aaronschachter: Your testing instructions mention changing this `null` to a value in order to test -- should this be a config variable as well? Or could the testing instructions could have been also to change the `DEFAULT_INTERVAL_MS` in Profiler (which already suggested could be optionally defined as a config variable)

aaronschachter:
Also not sure if this is expected or if I've missed anything, after about 15 minutes of activity I see these errors locally...

```
10:03:45 PM web.1 |  info: Upload Stream error: Failed to upload a part to S3: {"message":"Connection timed out after 120000ms","code":"NetworkingError","time":"2017-04-04T05:03:44.220Z","region":"us-east-1","hostname":"s3.amazonaws.com","retryable":true}
10:03:45 PM web.1 |  info: Snapshot error: Failed to upload a part to S3: {"message":"Connection timed out after 120000ms","code":"NetworkingError","time":"2017-04-04T05:03:44.220Z","region":"us-east-1","hostname":"s3.amazonaws.com","retryable":true}
```

lib/tools/profiler.js
rapala61: I prefer to expose ways to override in the constructor, these are just internal defaults. Overriding these defaults is an implementation dependent detail. I think we can easily also add them to the `.env.example` but use them where we instantiate the profiler, rather than in the profiler itself.  I'm not currently exposing NAME_PREFIX, so that is a change I should make. Does that makes sense?

server.js
rapala61: I agree, this would have been a perfect place for an env variable. I will change that and update the instructions.

server.js
rapala61: I agree, will change.

rapala61:
@aaronschachter I wasn't able to reproduce the error you mentioned using the S3 credentials in the Gambit S3 secure document stores in LastPass. I tested enabling the profiler and setting the interval to every 30 minutes, this is my log:

```
rafacode@rafamp61
 in   ~/Sites/ds/gambit $ heroku local
[OKAY] Loaded ENV .env File as KEY=VALUE Format
8:45:35 AM web.1 |  debug: Master called
8:45:35 AM web.1 |  info: Started worker 1
8:45:36 AM web.1 |  info: HEAP SNAPSHOT PROFILER IS ENABLED.
8:45:36 AM web.1 |  info: conn.readyState:1
8:45:36 AM web.1 |  debug: Campaign.lookupByIds:3539,2299,819,46
8:45:37 AM web.1 |  debug: Campaign.lookupByIds found 4 campaigns
8:45:37 AM web.1 |  info: Checking messaging groups for Campaign 46
8:45:37 AM web.1 |  info: Checking messaging groups for Campaign 819
8:45:37 AM web.1 |  info: Checking messaging groups for Campaign 3539
8:45:37 AM web.1 |  info: Checking messaging groups for Campaign 2299
8:45:37 AM web.1 |  info: Gambit is listening on port:5000 env:thor.
9:15:36 AM web.1 |  info: snapshotCreated has been emitted
9:15:58 AM web.1 |  debug: Snapshot upload successful. File name: v8.heap.snapshot_worker1_1491311736023.heapsnapshot
9:45:37 AM web.1 |  info: snapshotCreated has been emitted
9:46:08 AM web.1 |  debug: Snapshot upload successful. File name: v8.heap.snapshot_worker1_1491313537017.heapsnapshot
```

server.js
```diff
@@ -59,6 +59,63 @@ function startWorker(id) {
...
+   * Profiler setup
+   */
+
+  if (process.env.V8_PROFILER_ENABLED &&
+      process.env.AWS_ACCESS_KEY_ID &&
+      process.env.AWS_SECRET_ACCESS_KEY) {
+    logger.info('HEAP SNAPSHOT PROFILER IS ENABLED.');
+
+    const profiler = new Profiler({
+      interval: process.env.V8_PROFILER_DEFAULT_INTERVAL_MS,
```
aaronschachter: What about keeping the default values here in case the env variable doesn't exist?
```
process.env.V8_PROFILER_DEFAULT_INTERVAL_MS || 1800000 
process.env. V8_PROFILER_NAME_PREFIX || 'v8.heap.snapshot' 
```
Or updating the `.env.example` with values.. but my preference would be setting defaults and not requiring a config var to be set.

aaronschachter:
Forgot to update my `.env` which resulted in a lot of `error: uncaughtException: Cannot read property 'replace' of undefined` when testing locally --- which is where those default values come in handy....

rapala61:
@aaronschachter I see what you mean. The way I would go about it, would be to fix this at the constructor level, like this:
- Before: `this._sanitize(options.namePrefix) || NAME_PREFIX;`
- After: `this._sanitize(options.namePrefix || NAME_PREFIX);`

My thinking behind this is that i don't want an implementation detail to break the profiler, thats why im setting defaults internally, although as you pointed out, that was faulty.

My question now is, should I just let a wrong implementation make the profiler die like what you experienced? I think this is a more of a design pattern paradigm question than a bug, but I'm open to learn and change it.

aaronschachter:
Our general approach with Gambit has been to just log an error every time something fails that shouldn't take down the house with it. We could/should send a stathat alert if there's a profiler error-- one reason I haven't completely closed out #714 is we could do more with their alerts to let us know via email or slack when something's wrong.

For my scenario, where I didn't have the `.env` files set, that's where I'm suggesting we have backup defaults -- so if someone fresh to the project wanted to run Profiler, they'd essentially only need to set the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `V8_PROFILER_ENABLED`. I suggested making the other settings as config variables so we can tweak without code changes (per your initial testing instructions of modifying the param to the constructor).

lib/tools/profiler.js
```diff
@@ -0,0 +1,118 @@
...
+  * It generates the name of the snapshot.
+  * The snapshot's name follows the next pattern:
+  *    <PREFIX>_<TAG>_<EPOCH TIMESTAMP>
+  *    Example: v8.heap.snapshot_iamatag_1491242331638
+  *
+  * @param {Number} - optional timestamp
+  *
+  * @returns {String}
+  */
+  _getName(timestamp) {
```
sergii-tkachenko: No need to prefix functions with `_`: https://github.com/airbnb/javascript#naming--leading-underscore

lib/tools/uploader.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const S3 = require('aws-sdk/clients/s3');
+const s3Stream = require('s3-upload-stream');
+
+/**
+ * The Uploader class.
+ */
+class Uploader {
```
sergii-tkachenko: Might make sense to make naming more explicit, e.g. `S3Uploader`.

server.js
sergii-tkachenko: +1

rapala61:
@aaronschachter I'll set the defaults like you suggested. Does this mean that i should remove the defaults inside profiler?

rapala61:
@aaronschachter @sergii-tkachenko updated all the things! please review, im ready.

aaronschachter:
Looks good here 👍

.env.example
```diff
@@ -3,6 +3,13 @@ LOGGING_LEVEL=info
 
 WEB_CONCURRENCY=1
 
+V8_PROFILER_ENABLED=
```
sergii-tkachenko: I'd also recommend to prefix env variables with `GAMBIT_`, in case we wanted to expose them through normal env vars. So they don't conflict with other projects that may want to use, for example, AWS bucket.

lib/tools/profiler.js
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @params {Object} options - Object with the profiler configuration options
+   *                  interval: The amount in milliseconds to wait before taking snapshots
+   *                  tag: custom identifier
+   *                  extension: The extension of the file. It must be of the type:
+   *                             .cpuprofile, .heapsnapshot, .heapprofile, or .heaptimeline
+   */
+  constructor(options) {
+    super();
+    this.extension = this.sanitize(options.extension || '');
```
sergii-tkachenko: Doesn't make sense to sanitize empty string.

rapala61: I think it makes sense to sanitize the extension option, the empty string is just a last resort fallback to not trigger an exception as undefined would not have a function named `replace`:
```
  sanitize(string) {
    return string.replace(/[&@:,$=+?;\s^`><{}[\]#%~|\\"]/ig, ''); // here
  }
```

.env.example
sergii-tkachenko: Nevermind, I thought all other variables in Gamibt are prefixed with `GAMBIT_` too.

lib/tools/profiler.js
sergii-tkachenko: Yes, but it's possible to avoid unnecessary sanitation of `''` fallback:

```
this.extension = this.sanitize(options.extension) || '';
```

sergii-tkachenko: Per discussion with Rafa, this vars are not required.

rapala61: The tradeoff here is: check for this empty string in remote cases VS trying to upload a file and getting it rejected because of an unsafe name

Issue #849, opened 2017-04-10
Keep track of processing time for incoming request, send timeout response


aaronschachter:
_FEATURE OVERVIEW_
-------------------
#### User Story
As an O'Doyle dev -- I want to know if a request is taking longer than 30 seconds and receive a 503 back.

#### Additional Information (optional)
https://devcenter.heroku.com/articles/request-timeout
> When a timeout is detected the router will return a customizable error page to the client and an H12 error is emitted to your application logs. While the router has returned a response to the client, your application will not know that the request it is processing has reached a time-out, and your application will continue to work on the request. To avoid this situation Heroku recommends setting a timeout within your application and keeping the value well under 30 seconds, such as 10 or 15 seconds. Unlike the routing timeout, these timers will begin when the request begins being processed by your application. You can read more about this below in Timeout behavior.

#### Why This Matters
Blink will rely this response to retry a request if a 503 is sent back.

#### Current Workaround w/o Feature (if applicable)
N/A

aaronschachter:
This seems promising: https://www.npmjs.com/package/connect-timeout

aaronschachter:
Tried out of the new [TODO](https://github.com/DoSomething/gambit/blob/9d17bad0017f0e01f0daade02e6a759ab40ce8d9/app/routes/chatbot.js#L108)'s to use Promises to only return a Gambit 200 back to Blink upon getting a successful Mobile Commons **POST** `/profile_update`response in this commit -- https://github.com/DoSomething/gambit/commit/d43c4d725363cdad85b58528c9ae2b8da05d80a9. Check out the timestamps for waiting for the MC response back:

```
4:26:40 PM web.1 |  debug: helpers.replacePhoenixCampaignVars campaignId:2900
4:26:40 PM web.1 |  verbose: saved user:5547be89469c64ec7d8b518d current_campaign:2900
4:26:40 PM web.1 |  debug: mobilecommons.executePost:profile_update
4:26:40 PM web.1 |  debug: stathat response:200 postStat:'gambit-dev - phoenix: GET signups 200'
4:27:09 PM web.1 |  debug: res.text
4:27:10 PM web.1 |  debug:  
4:27:10 PM web.1 |  { success: 'true' }
4:27:10 PM web.1 |  error: Request timed out after 15 seconds.
4:27:11 PM web.1 |  debug: stathat response:200 postStat:'gambit-dev - 504: Request timed out after 15 seconds.'
```
[...]

Issue #824, opened 2017-03-21
Refactoring Gambit: Splitting out Chatbot Functions


jamjensen:
_FEATURE OVERVIEW_
-------------------
#### User Story
- The main chatbot router handles the majority of Gambit requests. We plan to split it out into smaller, middleware functions (such as: GetCampaigns, GetKeywords, etc.) so that we can add test coverage to understand gambit performance and so that we reduce the complexity of the main `chatbot` router.

#### Additional Information (optional)

- https://dosomething.slack.com/archives/C2C8NLNAY/p1490115138653699

PR #852, opened 2017-04-12
Check for timeouts in signups router

aaronschachter:
#### What's this PR do?
* Introduces [`connect-timeout`](https://www.npmjs.com/package/connect-timeout) and adds it at top level middleware in `server.js` to use for every incoming request.

* Splits up the `signups` router **POST** `/` into middleware functions, checking for `req.timedout` per the [module docs](https://github.com/expressjs/timeout/tree/v1.8.0#api). Keeps the order of operations the same:
    * Find the Signup by ID
    * Ensure Signup Campaign is active
    * Ensure Signup Campaign has published keywords
    * Ensure Signup User has a mobile number
    * Render the External Signup Message for the Campaign
    * Post Mobile Commons profile update to deliver the message via SMS to User

* Adds an optional `GAMBIT_TIMEOUT_NUM_SECONDS` config variable, default timeout at 15s. If reached, Gambit responds with a 504 response:

```
{
  "error": {
    "code": 504,
    "message": "Request timed out after 10 seconds."
  }
}
```

I was testing this locally by using XCode's Network Link Conditioner to simulate a slow connection, and verified I'd get 504's back. I'm not exactly sure how we can write an automated test to simulate this, Artsy has an [example of how to do so in Rails](http://artsy.github.io/blog/2012/11/15/how-to-monitor-503s-and-timeout-on-heroku/).  One idea is to use [Nock to delay a response](https://github.com/node-nock/nock#delay-the-response) and write a test using supertest, like we do in the [MGAPI](https://github.com/DoSomething/messaging-groups-api/blob/2.0.0/test/api.js).

Next up will be adding the checks into the `chatbot` router, but it warrants beginning work on #824 simultaneously.

#### Relevant tickets
#849, #824 

#### Checklist
- [x] Tested on staging.


sergiitk (approved):
This is amazing.

PR #853, opened 2017-04-12
Adds timedout check to Chatbot find/create user

aaronschachter:
#### What's this PR do?
Begins work on #824 and #849 by splitting out checks for finding/creating a User for a given phone number, and adding `req.timedout` checks per #852.

* Use existing user, create new user
* Test with local connectivity/timeout settings to verify request will timeout:
```
9:50:35 AM web.1 |  debug: User.lookup(mobile, [removed])
9:50:47 AM web.1 |  debug: northstar.Users.lookup success
9:50:47 AM web.1 |  error: Request timed out after 2 seconds.
```
Next up: splitting out loading Campaign and Signup into smaller functions. 

#### Relevant tickets
#824, #849, #852

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -70,6 +70,24 @@ router.use((req, res, next) => {
 });
 
 /**
+ * Check for required body parameters.
+ */
+router.use((req, res, next) => {
+  if (!req.body.phone) {
+    return helpers.sendUnproccessibleEntityResponse(res, 'Missing required phone.');
+  }
+  if (!(req.body.args || req.body.mms_image_url)) {
+    return helpers.sendUnproccessibleEntityResponse(res, 'Missing required args or mms_image_url.');
+  }
+  /* eslint-disable no-param-reassign */
+  req.incoming_message = req.body.args;
+  req.incoming_image_url = req.body.mms_image_url;
+  /* eslint-enable no-param-reassign */
+
+  return next();
+});
+
+/**
  * Posts to chatbot route will find or create a Northstar User for the given req.body.phone.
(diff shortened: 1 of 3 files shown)


app/routes/chatbot.js
```diff
@@ -44,58 +45,16 @@ function isCommand(incomingMessage, commandType) {
...
-    incomingMessage: scope.incoming_message,
-    keyword: scope.keyword,
-    mobileCommonsBroadcastId: scope.broadcast_id,
-    mobileCommonsMessageId: scope.body.message_id,
-    mobileCommonsProfileId: req.body.profile_id,
-  });
-
-  let configured = true;
-  // Check for required config variables.
+router.use((req, res, next) => {
```
sergii-tkachenko: A question: won't this middleware be executed on every request, despite actual route?

aaronschachter: I commented out my `MOBILECOMMONS_OIP_CHATBOT` variable and confirmed this function doesn't execute when posting to the Signups endpoint.

aaronschachter: From https://expressjs.com/en/api.html#router:
> The top-level express object has a Router() method that creates a new router object.

We declare `express.Router()` in each route, which seems to be reason why they are separate. 

https://github.com/DoSomething/gambit/blob/3.10.0/app/routes/chatbot.js#L7
https://github.com/DoSomething/gambit/blob/3.10.0/app/routes/signups.js#L4

sergii-tkachenko: Perfect.

sergiitk (approved):
👍

PR #854, opened 2017-04-13
Chatbot middleware to determine current Campaign 

aaronschachter:
#### What's this PR do?
* Splits out the code that checks for the current Campaign into middleware functions, maintaining the same order:
   * If a `broadcast_id` was sent, query Contentful Broadcasts to set a `req.campaignId` (and send the Declined Message if User sent a No response)
   * If a `keyword` was sent, query Contentful Keywords to set a `req.campaignId`
   * Else we should have a current Campaign ID set on the User document already.

* Adds timeout checks to the Contentful queries

* Verify expected behavior for the three scenarios listed above

Next up for splitting into middleware (and adding timeout checks):
* Loading the Campaign via Phoenix API
* Finding/creating the Signup via Phoenix API 

#### Relevant tickets
#853, #849, #824 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -311,6 +311,7 @@ router.post('/', (req, res) => {
       }
 
       if (scope.broadcast_id) {
+        // TODO: Add new parameter for broadcast_id to Signup post instead of saving here.
         scope.signup.broadcast_id = scope.broadcast_id;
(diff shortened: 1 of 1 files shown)


sergiitk (approved):
👍

PR #855, opened 2017-04-13
Cleanup

aaronschachter:
#### What's this PR do?
* Removes code for handling broadcast declined that never gets called, missed this in #854 
* Implements `helpers.sendUnproccessibleEntityResponse` in Signups to DRY
* Removes unused `UnproccessibleEntityError` and `Promise` vars -- Bluebird was required for filtered catch, but Signups simply returns the `UnproccessibleEntityResponse` instead of throwing a `UnproccessibleEntityError` to catch later.

👀 

Cleaning house 🏡 

#### Checklist
- [x] Tested on staging.


sergiitk (commented):
👍

PR #861, opened 2017-04-17
Cleanup Signup error responses, deprecate NotFoundError class

aaronschachter:
#### What's this PR do?
* Returns an error status code if we have one in the **POST** `/signups` response by using `helpers.sendErrorResponse`
   * Middleware refactory for Signups took place in #852, `helpers.sendErrorResponse` was introduced in #853, which forgot to go back and cleanup our Signup errors

* Deprecates the `NotFoundError` class. The custom `app/exceptions` were put in place to use filtered catches by our monster Promise chains, which we're looking to deprecate by splitting into smaller middleware functions (#824). If we're handling an error as it comes up, we can use `helpers.sendErrorResponse` to check for a HTTP status and send it back with the Gambit response.

* **POST** `/signups` with a Signup id that doesn't exist. Confirm a 404 is sent back, not a 500
* **POST** `/campaigns/:id/message` with a Campaign id that doesn't exist. Confirm a 404 is sent back, not a 500

Next up is closing out #849 for good by introducing timedout checks into the Campaigns endpoint. Noticed some cleanup we could do in Signups when posting to a Signup ID that didn't exist (got a 500 back instead of expected 404)

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/campaigns.js
+++ b/app/routes/campaigns.js
@@ -6,7 +6,6 @@ const Promise = require('bluebird');
 const logger = require('winston');
 
 const ClosedCampaignError = require('../exceptions/ClosedCampaignError');
-const NotFoundError = require('../exceptions/NotFoundError');
 const UnprocessibleEntityError = require('../exceptions/UnprocessibleEntityError');
 
 const contentful = require('../../lib/contentful');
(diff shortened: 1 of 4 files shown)


app/routes/campaigns.js
```diff
@@ -115,9 +107,8 @@ router.get('/:id', (req, res) => {
     })
     // TODO: Refactor helpers.sendResponse to accept an error and know the codes based on custom
     // error class, to DRY.
```
aaronschachter: This is step 1 of 2 for completing this `TODO` -- next up would be using the `helpers.sendUnproccessibleEntityProcess` (introduced in #853) in a middleware function checking for required parameters.

sergiitk (approved):
👍

PR #863, opened 2017-04-17
POST Campaign Message middleware + timeout check

aaronschachter:
#### What's this PR do?
Splits up **POST** `/campaigns/:id/message` into middleware and adds `req.timedout` checks.

Using Thor:

* 200: **POST** `/campaigns/7/message` with type `scheduled_relative_to_signup_date`

* 422: **POST** `/campaigns/7/message` with type `scheduled_relative_to_reportback_date` -- We don't have a default `scheduled_relative_to_reportback_date` set.

* 404: **POST** `/campaigns/723423/message` with type `scheduled_relative_to_signup_date`

* 422: **POST** `/campaigns/362/message`-- Campaign is closed

* 422: **POST** `/campaigns/62/message` -- Campaign is not running on Gambit (no keywords)

Technically we shouldn't close #504 without adding the `req.timedout` checks into the GET campaigns endpoints, but those are only used internally by Quicksilver to cache Campaigns, and good ol' Slothbot for internal staff usage... 

#### Relevant tickets
#504 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/exceptions/UnprocessibleEntityError.js
+++ b/app/exceptions/UnprocessibleEntityError.js
@@ -1,5 +1,12 @@
 'use strict';
 
-class UnprocessibleEntityError extends Error {}
+class UnprocessibleEntityError extends Error {
+
+  constructor(message) {
+    super(message);
+    this.status = 422;
+  }
+
+}
 
 module.exports = UnprocessibleEntityError;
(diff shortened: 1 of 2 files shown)


sergii-tkachenko:
I'm with @aaronschachter on error codes. 404 means page not found. 422 means the request is understood, but not valid:

https://tools.ietf.org/html/rfc4918#section-11.2

aaronschachter:
Agreed -- I'm not into the 404, because the resource exists.

PR #875, opened 2017-04-28
Issue 866/multi word yes responses

### LINKED ISSUE #866: Multi-word acceptable YES responses do not pass the isYesResponse test in lib/helpers
_BUG_
-------------------
#### Current Behavior
- Found while adding tests for `lib/helpers`
- `helpers.isYesResponse('i can').should.be.true;` returns `false` even though "I CAN" is an acceptable "Yes" response.

#### Desired Behavior
- `helpers.isYesResponse('i can').should.be.true;` should return `true`.

#### Steps to Replicate
- run the tests, test should fail.

#### Why This Matters
- Gambit can potentially misunderstand a reply that should be accepted as a Yes response.

rapala61:
#### What's this PR do?
Refactored the `isYesResponse` method in our `lib/helpers` file.
Adds test for `isYesResponse`.

- Install locally
- Make sure your GAMBIT_YES_RESPONSES env variable is populated with the latest accepted responses in production
- Run `npm run all-tests`
- All tests should pass

Accepted responses that where longer than 1 word were being ignored, example: "i can, hell yea", this change fixes that.

#### Relevant tickets
Fixes #866 

#### Checklist

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -60,11 +60,13 @@ module.exports.getFirstWord = function (message) {
 };
 
 module.exports.isYesResponse = function (message) {
-  const trimmed = this.getFirstWord(message.toLowerCase().trim());
+  const trimmedMsg = message.toLowerCase().trim();
   const yesResponses = process.env.GAMBIT_YES_RESPONSES ?
     process.env.GAMBIT_YES_RESPONSES.toLowerCase().split(',') : ['yes'];
 
-  return yesResponses.indexOf(trimmed) >= 0;
+  const matchRegex = new RegExp(`^(${yesResponses.join('|')})$`);
+
+  return !!trimmedMsg.match(matchRegex);
 };
 
 module.exports.hasLetters = function (message) {
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ node_modules
 #code coverage directories
 .nyc_output
 coverage
+
+#local nvm configuration
+.nvmrc
--- a/package.json
+++ b/package.json
@@ -8,9 +8,9 @@
     "url": "git+https://github.com/DoSomething/gambit.git"
   },
   "scripts": {
-    "test": "NODE_ENV=test ava",
+    "test": "NODE_ENV=test ava --serial",
     "all-tests": "npm run lint && npm run coverage",
-    "coverage": "NODE_ENV=test nyc --all ava",
+    "coverage": "NODE_ENV=test nyc --all ava --serial",
     "html-coverage-report": "nyc report --reporter=html",
     "publish-coverage-report": "nyc report --reporter=lcov && codecov",
     "start": "node server",
@@ -66,6 +66,7 @@
   "devDependencies": {
     "@dosomething/eslint-config": "^2.0.0",
     "ava": "^0.19.1",
+    "bluebird": "^3.5.0",
     "chai": "^3.5.0",
     "codecov": "^2.1.0",
     "dotenv": "^4.0.0",
@@ -73,6 +74,7 @@
     "eslint-plugin-ava": "^4.2.0",
     "node-mocks-http": "^1.6.1",
     "nyc": "^10.2.0",
-    "sinon": "^2.1.0"
+    "sinon": "^2.1.0",
+    "sinon-chai": "^2.9.0"
   }
 }
--- a/test/lib/helpers.test.js
+++ b/test/lib/helpers.test.js
@@ -1,44 +1,65 @@
 'use strict';
 
 require('dotenv').config();
+const Promise = require('bluebird');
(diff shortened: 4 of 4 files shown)


aaronschachter:
Not sure if I'm missing something, but I've pulled down this branch and ran `npm install`.

`npm run all-tests` is throwing a bunch of errors for the linter, this is happening on the develop branch as well:
```
/Users/aschachter/Development/gambit/app/exceptions/ClosedCampaignError.js
  1:1  error  Definition for rule 'react/style-prop-object' was not found         react/style-prop-object
  1:1  error  Definition for rule 'react/jsx-wrap-multilines' was not found       react/jsx-wrap-multilines
```

Running `npm test` on this branch throws a few errors for:
```
    [TypeError: Attempted to wrap fetchKeywordsForCampaignId which is already wrapped]
```

rapala61:
Tried deleting node_modules and re running npm install?

aaronschachter:
@rapala61 That fixed the linting errors, but I'm still getting the `TypeError` on this branch:

```
  helpers › beforeEach for isValidReportbackText
  /Users/aschachter/Development/gambit/node_modules/sinon/lib/sinon/util/core/wrap-method.js:39

  Error thrown in test

  Error:

    [TypeError: Attempted to wrap fetchKeywordsForCampaignId which is already wrapped]
```

rapala61:
Strange. Does it happen after a particularly failed test?

aaronschachter:
Ahh, yes, it does fail on `isYesResponse` first -- because I didn't have my `GAMBIT_YES_RESPONSES` set in my `.env`... which brings up #876 and config variables + test coverage.

Can we refactor these tests to test what's defined in the `.env` instead of hardcoded values? It kind of defeats the purpose of config values if we need to update the test code accordingly. I could see an argument for keeping those values in code, but for something like Yes Responses, where I can see additions/removals more often -- it's definitely convenient to make the config changes vs code changes....

rapala61:
@aaronschachter I think it would help me to define what it is that we are testing. In the case of the isYesResponse function, I am feeding x,y,z data to the function and assert that some of these are acceptable and some are not. How do I decide to assert false or true? I go by what the production acceptable responses env variable value is. I am not hard coding what the actual valid responses are, that is dynamic with the env variable, what is hard coded, are the test cases, which is a total valid way of testing in my opinion, its like feeding garbage values to a function that is supposed to return a response.

rapala61:
Having your tests fail because you have no GAMBIT_YES_RESPONSES env variable is a reasonable reason why it fails.

aaronschachter:
Makes sense... but what about the scenario for the #876 where we want to change the `CMD_MEMBER_SUPPORT` value one day, we'd then need to update the test accordingly to test the `true` case? That one's a realistic scenario where we'd want to change it, it's only Q because the QUESTION keyword is already defined on a different Mobile Commons Campaign (which will one day be deprecated).

lib/helpers.js
```diff
@@ -60,11 +60,13 @@ module.exports.getFirstWord = function (message) {
 };
 
 module.exports.isYesResponse = function (message) {
-  const trimmed = this.getFirstWord(message.toLowerCase().trim());
+  const trimmedMsg = message.toLowerCase().trim();
   const yesResponses = process.env.GAMBIT_YES_RESPONSES ?
     process.env.GAMBIT_YES_RESPONSES.toLowerCase().split(',') : ['yes'];
 
-  return yesResponses.indexOf(trimmed) >= 0;
+  const matchRegex = new RegExp(`^(${yesResponses.join('|')})$`);
```
aaronschachter: Minor, but it might be nice to leave a comment to translate this RegExp into readable. Will this find a yes response anywhere in the string? e.g. "I'm thinking about it and the answer is YES"

rapala61: The change attempts to behave as close as possible to the original business logic (which was to only contemplate the beginning of the text) while fixing the bug. In this case "I'm thinking about it and the answer is YES" is not a Yes response now, or before the code changes. If the business logic has changed, we should make a separate issue and address that.

aaronschachter: Ok, just checking it's still the same. It'd be nice to have the readable comment, `// Check if string starts with a Yes response`, but not necessary for merge.

rapala61:
#876 is also testing the business logic. If we want to deliberately be able to change it at our leisure without considering it a potential bug, then I'm perfectly on board with not testing for an specific value, and this is true with any env variable. Also, I have contemplated testing for existence of env variables as a separate test in the future, I think there is also some comments about this in Gambit, but this will also need refactoring, which will happen anyways to give way for functional tests.

lib/helpers.js
rapala61: I'll add the comment in the next testing pull requests.

PR #879, opened 2017-05-01
Send 404 if keyword not found

### LINKED ISSUE #874: Keyword not found should return 404
_BUG_
-------------------
#### Current Behavior
```
curl -X POST \
  http://ds-mdata-responder-staging.herokuapp.com/api/v1/webhooks/gambit-chatbot-mdata \
  -H '<AUTH HEADER>' \
  -H 'content-type: application/json' \
  -d '{
  "keyword": "SHARE",
  "phone": "1<PHONE NUMBER>",
  "message_id": "841415468",
  "profile_id": "167181555"
}'
```

2017-04-26T19:01:31.617289+00:00 heroku[router]: at=info method=POST path="/v1/chatbot" host=ds-mdata-responder.herokuapp.com request_id=afbd6544-b75b-406c-a932-f5f05a69da4d fwd="54.159.150.62" dyno=web.1 connect=0ms service=40ms **status=404** bytes=273 protocol=http

#### Desired Behavior
status=422

#### Why This Matters
So blink won't retry it 100 times.

aaronschachter:
This reverts commit b45a5280a954518e8b631b0f6bd69d9517365449.

#### What's this PR do?
Sends a 404 if a keyword is not found in Contentful, instead of a 422. 

Pass a keyword that isn't in Contentful to a POST `chatbot` request and verify a 404 is returned.

Discussed with @sergii-tkachenko and @rapala61 today -- we do want Blink to retry when a keyword isn't found. It's likely because either the keyword needs to be added to Contentful, or it needs to be removed from the mData.

#### Relevant tickets
Fixes #874 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -296,7 +296,7 @@ router.use((req, res, next) => {
         return helpers.sendTimeoutResponse(res);
       }
       if (!keyword) {
-        return helpers.sendUnproccessibleEntityResponse(res, `Keyword ${req.keyword} not found.`);
+        return helpers.sendResponse(res, 404, `Keyword ${req.keyword} not found.`);
       }
       if (keyword.fields.environment !== process.env.NODE_ENV) {
         const msg = `Keyword ${req.keyword} environment error: defined as ${keyword.environment} ` +


sergiitk (commented):
👍

PR #886, opened 2017-05-03
Bug fix in POST Signups 

### LINKED ISSUE #885: Signups error: res.status is not a function
_BUG_
-------------------
#### Current Behavior
Seeing 500 HTTP errors in New Relic with `gambitErrorMessage` with value`res.status is not a function`. This is 🙅 

Refs #882

#### Desired Behavior
`res.status` should totes be a function

#### Steps to Replicate
Not sure yet

#### Why This Matters
External Signup messages are not being delivered to what seems to be just a handful of users (only seeing 2 instances of this over past 6 hours)

https://rpm.newrelic.com/accounts/108038/applications/25522350/traced_errors/20927f86-2f95-11e7-bb42-0242ac11000b_0_878

aaronschachter:
#### What's this PR do?
* Fixes #885 by passing our Express `res` variable to a `sendUnproccessibleEntityResponse` call.
* Adds New Relic custom parameters to help debug errors logged in New Relic

This was happening for the scenario when Quicksilver is posting Signups for Campaigns that no longer have keywords published. I've unpublished SHOWERBOT to verify I get a 422 back instead a 500 with `res.status is not a function`

The good news here is we didn't want to send Extneral Signup Messages in this case anyway, because the DS Campaign is no longer available on Gambit. Minorly bad news is that it means Quicksilver cache should be restarted? cc @jamjensen @sergii-tkachenko 

#### Relevant tickets
Fixes #885 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/signups.js
+++ b/app/routes/signups.js
@@ -49,7 +49,9 @@ router.use((req, res, next) => {
       if (req.timedout) {
         return helpers.sendTimeoutResponse(res);
       }
+
       req.signup = signup; // eslint-disable-line no-param-reassign
+
       return next();
     })
     .catch(err => helpers.sendErrorResponse(res, err));
@@ -128,7 +130,9 @@ router.use((req, res, next) => {
       if (req.timedout) {
         return helpers.sendTimeoutResponse(res);
       }
+
       req.signupMessage = helpers.addSenderPrefix(message); // eslint-disable-line no-param-reassign
+
       return next();
     })
     .catch(err => helpers.sendErrorResponse(res, err));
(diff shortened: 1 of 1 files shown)


app/routes/signups.js
```diff
@@ -81,10 +89,12 @@ router.use((req, res, next) => {
       if (req.timedout) {
         return helpers.sendTimeoutResponse(res);
       }
+
       if (keywords.length === 0) {
         const msg = `Campaign ${req.campaign.id} does not have any Gambit keywords.`;
-        return helpers.sendUnproccessibleEntityResponse(msg);
+        return helpers.sendUnproccessibleEntityResponse(res, msg);
```
aaronschachter: The culprit.

rapala61: 🐛

rapala61 (approved):

PR #887, opened 2017-05-03
Trim Reportback text to avoid Phoenix API errors

### LINKED ISSUE #881: POST Reportback fails if caption > 255
_BUG_
-------------------
#### Current Behavior
If a `reportback_submission` contains a caption with length greater than 255 characters and is posted to the Phoenix API as a Reportback,  Gambit returns a 500 error with message: `Signup.postDraftReportbackSubmission error:API response is false.`

In Phoenix, the Reportback is saved without an Item. Phoenix logs show: `PDOException: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'caption' at row 1 in ...`

#### Desired Behavior
- The Reportback Item should be saved, with a truncated caption.

#### Why This Matters
- Missing Reportback images  🙅 
- Now that Blink will retry requests that return a 500 error, our Gambit error rate will skyrocket when we get errors like these.

screen shot 2017-05-02 at 9 45 31 am

aaronschachter:
#### What's this PR do?
Trims data we send for Reportback `caption` and `why_participated` to avoid Phoenix API errors (which result in Gambit returning an error response)

Fire up Lorem Ipsum generator and send through values for `caption` and `why_participated` that exceed 255 characters, verify reportback is successfully submitted (and long values are trimmed)

I'm having trouble finding the exact limit of `why_participated`. [The `length` value defined as `'2048'` for the `why_participated` field with type `'text'` in Phoenix doesn't seem correct](https://github.com/DoSomething/phoenix/blob/28d1ed6a5b8c8031c54f96244daee257c0e6398c/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install#L60), as I was able to pass a `why_participated` with 2049 characters (`'2048'` probably shouldn't be a string...)

Decided to share the 255 character limit for `why_participated` to keep the code simple for now.

#### Relevant tickets
Fixes #881 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -493,7 +493,7 @@ router.post('/', (req, res, next) => {
       return continueConversationWithMessageType(req, res, 'invalid_caption');
     }
 
-    draft.caption = input;
+    draft.caption = helpers.trimReportbackText(input);
 
     return draft.save()
       .then(() => {
@@ -516,7 +516,7 @@ router.post('/', (req, res, next) => {
       return continueConversationWithMessageType(req, res, 'invalid_why_participated');
     }
 
-    draft.why_participated = input;
+    draft.why_participated = helpers.trimReportbackText(input);
 
     return draft.save()
       .then(() => next())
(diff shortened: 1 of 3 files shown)


rapala61:
👍

PR #890, opened 2017-05-05
Contentful tests 2

rapala61:
#### What's this PR do?
Continued implementing tests for the `lib/contentful` module.
Some moderate refactoring involved.

- clone Gambit
- Run:
  - `npm install`
  - `npm run all-tests`

#### Checklist

### DIFF
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -9,8 +9,10 @@ const logger = require('winston');
 const config = require('../config/contentful');
 const helpers = require('./helpers');
 const UnprocessibleEntityError = require('../app/exceptions/UnprocessibleEntityError');
+const Builder = require('./contentfulQueryBuilder');
 
 const defaultCampaignId = config.defaultCampaignId;
+const ERROR_PREFIX = 'Contentful:';
 
 // The sequence we define peroperties here determines the order they appear in GET Gambit Campaigns
 // API response. Should match the sequence defined in Contentful, which is based on the order in
@@ -35,11 +37,14 @@ const contentfulCampaignFields = {
   campaign_closed: 'campaignClosedMessage',
 };
 
+module.exports.getQueryBuilder = function getQueryBuilder() {
+  return new Builder();
+};
+
 /**
  * Setup.
  */
 let client;
-
 
 /**
  * createNewClient
@@ -52,7 +57,7 @@ module.exports.createNewClient = function createNewClient() {
   } catch (err) {
     // TODO: Log this. Send to Stathat, but alert developers that we're in trouble.
     // @see https://github.com/DoSomething/gambit/issues/714#issuecomment-259838498
-    logger.error(`contentful error:${err.message}`);
+    logger.error(exports.contentfulError(err));
   }
   return client;
 };
@@ -74,9 +79,16 @@ module.exports.getClient = function getClient() {
  */
 module.exports.contentfulError = function contentfulError(error) {
   const scope = error;
-  scope.message = `Contentful: ${error.message}`;
+  scope.message = `${ERROR_PREFIX} ${error.message}`;
 
   return scope;
+};
+
+module.exports.getFirstItemFromArray = function getFirstItemFromArray(array) {
+  if (array.length === 0) {
+    return false;
+  }
+  return array[0];
 };
 
 /**
@@ -86,13 +98,7 @@ module.exports.fetchSingleEntry = function fetchSingleEntry(query) {
   return new Promise((resolve, reject) => {
     logger.debug(`contentful.fetchSingleEntry:${JSON.stringify(query)}`);
     return exports.getClient().getEntries(query)
-    .then((entries) => {
-      if (entries.items.length === 0) {
-        return resolve(false);
-      }
-
-      return resolve(entries.items[0]);
-    })
+    .then(entries => resolve(exports.getFirstItemFromArray(entries.items)))
     .catch(error => reject(exports.contentfulError(error)));
   });
 };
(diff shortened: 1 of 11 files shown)


lib/contentful.js
```diff
@@ -142,46 +151,54 @@ module.exports.fetchCampaignIdsWithKeywords = function () {
...
-  };
-  if (campaignId) {
-    query['fields.campaign.sys.contentType.sys.id'] = 'campaign';
-    query['fields.campaign.fields.campaignId'] = campaignId;
-  }
-
+    .env(process.env.NODE_ENV)
+    .custom({
+      'fields.campaign.sys.contentType.sys.id': 'campaign',
+      'fields.campaign.fields.campaignId': campaignId,
```
aaronschachter: This is where I'm a little bit on the fence on the `QueryBuilder` class abstraction. I can see how it's useful for the simple fetch queries like `fetchKeyword` or `fetchCampaign` to not need to know the inner workings of the Contentful content types, but the super consistent maybe OCD brain would prefer the Builder functions to match the field names, e.g. `campaignId` or `broadcastId` instead of `campaign`, `broadcast`, `environment` instead of `env`, `contentType` instead of `type`... 

Yet one still needs to know the inner workings of the Contentful content types to execute a complex query anyway with `custom`.  The devil's advocate question is does the `Builder` actually make things eaiser?

cc @sergii-tkachenko

lib/contentful.js
```diff
@@ -74,53 +79,57 @@ module.exports.getClient = function getClient() {
  */
 module.exports.contentfulError = function contentfulError(error) {
   const scope = error;
-  scope.message = `Contentful: ${error.message}`;
+  scope.message = `${ERROR_PREFIX} ${error.message}`;
 
   return scope;
 };
 
+module.exports.getFirstItemFromArray = function getFirstItemFromArray(array) {
```
aaronschachter: If we're abstracting this -- does it make sense to just add https://lodash.com/docs/4.17.4#head?

rapala61: I though about using underscore actually. I though It wasn't necessary to bring another library for this. If we start needing more methods, like the _.extend, I think it makes sense to just include it.

test/stubs/contentful/get-entries-keyword.json
```diff
@@ -0,0 +1,131 @@
...
+              "sys": {
+                "type": "Link",
+                "linkType": "ContentType",
+                "id": "campaign"
+              }
+            },
+            "locale": "en-US"
+          },
+          "fields": {
+            "campaignId": "2299"
```
aaronschachter: Not a biggie -- but would it be possible to define 2299 as a `testCampaignId` somewhere so it's not hardcoded in our various tests? We do that Phoenix JS [here](https://github.com/DoSomething/phoenix-js/blob/master/test/helpers/test-helper.js#L33)

rapala61: Yea, I can add that method in the stubs module, but this is the JSON of a response. Are you saying that you would like this JSON response to have a dynamic campaign id? I'm not sure how to do that.

lib/contentful.js
aaronschachter: `_.extend` does look like a clean way to avoid the disabling `no-param-reassign` lint rules in our middleware.... https://github.com/DoSomething/gambit/blob/c96bdbe67c6b10cdd377f6b182d8b7190f64912b/app/routes/chatbot.js#L241

My vote would be to include and use it instead of our own custom `getFirstItemFromArray` (that could be sense to move to `helpers` if we kept it instead of a different library.. but it sounds like we'd benefit from having the other underscore methods)

lib/contentful.js
rapala61: The abstraction is there for 2 main reasons. 1) To be able to write tests that can assert that a query has been built (behavior), not what fields the query has (implementation). 2) To structure the query building like you would do it in mongo for example, through chaining calls that add statements to the query.

I agree that there is always the need to "know" the inner workings of the contentful query fields for more complex queries, but this is true for any query builder I have tried, for every document oriented database driver. This is why I think having a method that allows you to just set a custom (complex) query for when we need to, its an acceptable trade-off.

rapala61: As far as the naming convention, if it adds clarity to the method to call it `campaignId` vs `campaign` then yes. To my brain, having a `campaignId` method, sort of lets me wonder if there is another way to add more things about the campaign or field, like `campaignName`, `contentTypeId`, etc. Calling it campaign, adds the benefit to override it to accept an object and send more specific data like `.campaign({id: 1, closed: false})`, which would be weird if it was named `.campaignId({id:1, closed: true})`. This is the reason I named it this way, but I'm totes open to what you guys think.

lib/contentful.js
rapala61: I agree, I'll pull underscore in and move the function to the helpers file.

aaronschachter: I was thinking we wouldn't need the function in helpers at all if we just called [`first`? ](http://underscorejs.org/#first)

test/stubs/contentful/get-entries-keyword.json
aaronschachter: Yeah, it's tougher to do that when we define these mocks as JSON files.. which definitely makes sense to do. Not a biggie.

lib/contentful.js
rapala61: Yeah! right 😄

lib/contentful.js
aaronschachter: Explanation for adding the Query Builder makes sense. I would prefer function names like `campaignId` though, in our routes we assign full objects like `req.user` and `req.campaign` so it doesn't feel consistent.

rapala61: Ok, cool, ill update that then. 👍

PR #891, opened 2017-05-05
Adds Dashbot analytics

aaronschachter:
#### What's this PR do?
Posts incoming and outgoing chatbot messages to Dashbot:
* Outgoing messages are tracked as the Gambit message type (e.g. `ask_photo`, `campaign_closed`, `menu_completed`)
* Incoming messages are tracked with `keyword:[keyword` if they contain a keyword, `(image)` if they contain an image, as well as whatever text a member texts in.

To test locally, add the Dashbot API key for Gambit Local and check its live feed to verify incoming/outgoing messages.
To test Thor, text in a Thor keyword and check the live feed of the Gambit Thor bot to verify incoming/outgoing messages.

Here are some example screenshots from testing with Gambit Local:

screen shot 2017-05-05 at 10 49 25 am
screen shot 2017-05-05 at 10 50 00 am
screen shot 2017-05-05 at 10 50 12 am

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,6 +37,8 @@ NEW_RELIC_LOGGING_LEVEL=info
 STATHAT_APP_NAME=gambit-dev
 STATHAT_EZ_KEY=puppet.sloth@dosomething.org
 
+DASHBOT_API_KEY=totallysecret
+
 GAMBIT_JR_API_BASEURI=http://dev-gambit-jr.pantheonsite.io/wp-json/wp/v2/
 DONORSCHOOSE_API_BASEURI=https://qa.donorschoose.org
 DONORSCHOOSE_API_KEY=totallysecret
--- a/app/models/User.js
+++ b/app/models/User.js
@@ -6,6 +6,7 @@
 const mongoose = require('mongoose');
 const Promise = require('bluebird');
 const logger = require('winston');
+const superagent = require('superagent');
 
 const helpers = rootRequire('lib/helpers');
 const mobilecommons = rootRequire('lib/mobilecommons');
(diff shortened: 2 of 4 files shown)


rapala61 (commented):
I think is a good idea to add the new env variables to the .env.example .

aaronschachter:
Oops, I did that at some point but forgot to commit. Good catch @rapala61 !

rapala61 (approved):

aaronschachter:
Updated @rapala61.

The outgoing messages are being tracked by message type, like we're doing for Stathat.

I did consider whether it'd be useful to track by Campaign ID too, so we could track outgoing messages by Campaign: e.g. `'campaign:5_ask_photo'` vs `'campaign:362_ask_photo'`, but thought overall it'd likely be more useful over time to see how many times we sent a message type across all campaigns.

I did send Dashbot a help ticket asking about getting [Template Tags](https://www.dashbot.io/sdk/template) to work with the Generic SDK, to group the outgoing messages by type if we did include the Campaign ID, but haven't heard back from them. My hunch is that it's not supported (mentioned it [here](https://github.com/DoSomething/gambit/commit/7da5225a351a19883e04ffcf0be3ad24f303ada6#commitcomment-22024455))

aaronschachter:
What's also a little tricky about these analytics and how we currently use so many keywords on the Master / FAQ campaigns in Mobile Commons -- is that we won't be tracking any incoming messages that takes the user out of the Gambit flow (e.g. asking "Are you a robot?") 

Once we get all traffic routed to Blink / Gambit -- this is something we'd likely want to do if we like Dashbot. it's one way we can view a member's message history with us besides the Mobile Commons Profile (and building our own) cc @jamjensen

aaronschachter:
Messages that don't have an mData attached (like our Reminder Messages) are currently not being tracked here either, as we aren't loading a full User model in that endpoint. Because the Reminder Message doesn't use `profile_update` to opt the member into the Chatbot Mobile Commons Campaign, we can't track the incoming replies they send to a Reminder Message either -- hence excluding it from our Dashbot analytics for now.

aaronschachter:
There's also https://botanalytics.co/ -- which also advertises as having [human takeover ](https://botanalytics.co/product/features), so it sounds worth taking a look at. I wasn't able to get the Human Takeover working with Dashbot as its only supported in [Slack and FB Messenger.](https://www.dashbot.io/sdk/pause)

aaronschachter:
One nice tweak to these analytics would be to track an outgoing message of `signup_broadcast` when we have a `broadcast_id` to better indicate our outgoing message vs incoming message count.

PR #892, opened 2017-05-08
Revisiting helpers test

### LINKED ISSUE #876: Replace hardcoded commands in helpers.test
_BUG_
-------------------
#### Current Behavior
- `npm test` fails when `GAMBIT_CMD_REPORTBACK` is not set to `start`, because it's hardcoded to check for 'start'

#### Desired Behavior
- The test should pass, checking for the `GAMBIT_CMD_REPORTBACK` variable

Refs #864

rapala61:
#### What's this PR do?
Refactores some tests for the `lib/helpers` module.

- clone Gambit
- Run:
  - `npm install`
  - `npm run all-tests`

Fixes #876 

#### Checklist


rapala61:
@aaronschachter I refactored the isYesResponse test, does this look better to you? I think this accomplishes a good balance of what we were talking about earlier.

aaronschachter:
I do think this makes more sense than hardcoding all of the Yes responses. I may be mistaken, but it t sounded like @sergii-tkachenko was referring to creating a new `helpers` function to get the config variable (or provide a default) within [`isYesResponse`](https://github.com/DoSomething/gambit/blob/develop/lib/helpers.js#L57).. I'm not sure if that'd be helpful? 

I think it'd be safe to close out #876 if we did something similar for getting the Start and Reportback commands (vs hardcoding the `'start'` and `'q'` strings in the tests)...

rapala61:
@aaronschachter The `isYesResponse` function already has a default in it. I feel better about this flavor of the test also. I will check how to implement the same pattern with the `start` and `q` command.

PR #903, opened 2017-05-18
Contestful tests 4

rapala61:
#### What's this PR do?
More refactoring of `lib/contentful`. Added more tests weeeee!.

- clone Gambit
- Run:
  - `npm install`
  - `npm run all-tests`

#### Checklist

### DIFF
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -201,68 +201,89 @@ module.exports.renderMessageForPhoenixCampaign = function (phoenixCampaign, msgT
   });
 };
 
-/**
- * Returns all rendered messages for given phoenixCampaign, using defaults when no override defined.
- */
-module.exports.renderAllMessagesForPhoenixCampaign = function (phoenixCampaign) {
-  logger.debug(`renderAllMessagesForPhoenixCampaign:${phoenixCampaign.id}`);
 
+/**
+ * Gets all fields for the campaign together with the default fields in the default campaign.
+ *
+ * @param  {type} campaignId
+ * @return {Promise}
+ */
+module.exports.getAllFieldsForCampaign = function getAllFieldsForCampaign(campaignId) {
   return new Promise((resolve, reject) => {
     const campaignFieldNames = Object.keys(config.campaignFields);
     const campaignContent = {};
-    const result = {};
-
-    return exports.fetchCampaign(phoenixCampaign.id)
+    const campaignFields = {};
+    return exports.fetchCampaign(campaignId)
       .then((contentfulCampaign) => {
         campaignContent.overrides = contentfulCampaign.fields;
-
         return exports.fetchCampaign(defaultCampaignId);
       })
       .then((defaultContentfulCampaign) => {
         campaignContent.defaults = defaultContentfulCampaign.fields;
-        const promises = [];
-
         campaignFieldNames.forEach((key) => {
           const fieldName = config.campaignFields[key];
-          result[fieldName] = {};
+          campaignFields[fieldName] = {};
 
           if (campaignContent.overrides[fieldName]) {
-            result[fieldName].override = true;
-            result[fieldName].raw = campaignContent.overrides[fieldName];
+            campaignFields[fieldName].override = true;
+            campaignFields[fieldName].raw = campaignContent.overrides[fieldName];
           } else {
-            result[fieldName].override = false;
-            result[fieldName].raw = campaignContent.defaults[fieldName];
-          }
-
-          promises.push(helpers.replacePhoenixCampaignVars(result[fieldName].raw, phoenixCampaign));
-        });
-
-        return Promise.all(promises);
-      })
-      .then((renderedMessages) => {
-        // The order of renderedMessages corresponds to the sequence we defined campaignFieldNames.
-        renderedMessages.forEach((renderedMessage, index) => {
-          // Store the machine name used in code (e.g. 'ask_caption').
-          const key = campaignFieldNames[index];
-          // Find the Contentful field name that corresponds to the machine name.
-          const fieldName = config.campaignFields[key];
-
-          if (result[fieldName]) {
-            result[fieldName].rendered = renderedMessage;
+            campaignFields[fieldName].override = false;
+            campaignFields[fieldName].raw = campaignContent.defaults[fieldName];
           }
         });
-
-        return resolve(result);
+        return resolve(campaignFields);
       })
       .catch(err => reject(err));
   });
 };
 
-// Utility functions
+/**
+ * Returns all rendered messages for given phoenixCampaign, using defaults when no override defined.
+ */
+module.exports.renderAllMessagesForPhoenixCampaign = function (phoenixCampaign) {
+  logger.debug(`renderAllMessagesForPhoenixCampaign:${phoenixCampaign.id}`);
+  return new Promise((resolve, reject) =>
+    // get all fields, overrides and default
+    exports.getAllFieldsForCampaign(phoenixCampaign.id)
+      .then((campaignFields) => { // eslint-disable-line arrow-body-style
+        /**
+         * Use Promise.map to create an array of promises to wait on,
+         * replacing the push->forEach->Promise.all boilerplate
+         */
+        return Promise.map(
+          /**
+           * Iterator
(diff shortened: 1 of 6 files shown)


lib/contentful.js
```diff
@@ -201,69 +201,90 @@ module.exports.renderMessageForPhoenixCampaign = function (phoenixCampaign, msgT
...
-  return Object.keys(config.campaignFields);
+/**
+ * Returns all rendered messages for given phoenixCampaign, using defaults when no override defined.
+ */
+module.exports.renderAllMessagesForPhoenixCampaign = function (phoenixCampaign) {
+  logger.debug(`renderAllMessagesForPhoenixCampaign:${phoenixCampaign.id}`);
+  return new Promise((resolve, reject) =>
+    // get all fields, overrides and default
+    exports.getAllFieldsForCampaign(phoenixCampaign.id)
+      .then((campaignFields) => { // eslint-disable-line arrow-body-style
```
aaronschachter: Why do we need to disable `arrow-body-style` here? would this be resolved by changing line 254 to `Promise.map(` instead of `return Promise.map(`?

rapala61: I wrestled with this idea as I was coding. It would, but it sacrifices code readability IMO. This is how it looks like without the `{}`.

```javascript
module.exports.renderAllMessagesForPhoenixCampaign = function (phoenixCampaign) {
  logger.debug(`renderAllMessagesForPhoenixCampaign:${phoenixCampaign.id}`);
  return new Promise((resolve, reject) =>
    // get all fields, overrides and default
    exports.getAllFieldsForCampaign(phoenixCampaign.id)

      /**
       * Use Promise.map to create an array of promises to wait on,
       * replacing the push->forEach->Promise.all boilerplate
       */
      .then(campaignFields => Promise.map(
          /**
           * Iterator
           */
          Object.keys(campaignFields),
          /**
           * Mapper fn: calls helpers.replacePhoenixCampaignVars for each campaign field raw message
           * NOTE Disabling arrow-body-style because we need to create a function scope to
           * remember `fieldName`. Disabling max-len for better readability
           */
          (fieldName) => { // eslint-disable-line arrow-body-style
            return helpers.replacePhoenixCampaignVars(campaignFields[fieldName].raw, phoenixCampaign) // eslint-disable-line max-len
              /**
               * Let's add the rendered message to the campaignFields object under the rendered prop
               */
              .then((msg) => {
                campaignFields[fieldName].rendered = msg; // eslint-disable-line no-param-reassign
              });
          })

        /**
         * When all promises have been fulfilled, we resolve our Promise with the campaignFields
         * rendered
         */
        .then(() => {
          resolve(campaignFields);
        })
        .catch(err => reject(err)))
      .catch(err => reject(err)));
};
```

test/stubs/contentful/get-entries-campaign-with-overrides.json
```diff
@@ -0,0 +1,40 @@
...
+            "type": "Link",
+            "linkType": "ContentType",
+            "id": "campaign"
+          }
+        },
+        "locale": "en-US"
+      },
+      "fields": {
+        "campaignId": "7",
+        "gambitSignupMenuMessage": "Great! Little moments can have a big impact. Complimenting someone can turn their whole day around.\n\nDo this: post encouraging notes in places that can trigger low self-esteem, like school bathrooms.\n\nThen, text START to share a photo of the messages you posted (and you'll be entered to win a $1500 scholarship)! \n\nBonus: Text back a photo of your posted note by 11:59PM tonight and you can enter to win a gift card of your choice! (Winners announced this week). ",
```
aaronschachter: The `text START` text will be one of those instances where if we ever change the Start Reportback Command config variable, these tests would break. Nothing we need to change for merging, but it does make me wonder if the stubbed objects should be returned by a function, where we can pass a parameter like the command reportback to render. (or the campaign ID to use as I've mentioned before).

There's definitely something to be said how it's convenient to just store the stub data as a json file too. Just food for thought 🍔 .

rapala61: @aaronschachter 
No test should be testing the data itself, just the existence of properties, and states of being empty or not, but not actual string comparisons, because of this same reason. The text you refer to is the actual value in contenful for this override, the `text START` is hardcoded as the string returned in this field, in contentful. This, is in my opinion, is a potential source for a bug since it could run out of sync with what we set in the env variable (although we rarely change this), but this is a different issue.

Any JSON stub I use, I think of it as stale data, so I don't test if the stubs particular string values are correct, but rather that the data structure is what I expect when stubbing functions and getting back the created objects, etc. I also use them to force the unit tests to follow a certain branch of the method, then I test for expected methods being called in that particular function branch.

Creating cyborg stubs really doesn't solve the problem that the actual values in them can and will run out of sync. If we just make them dynamic, we are going to run the risk of treating them like actual important sources of truth, relying on them and the fake sense of trust that will give us (sorry stubs, not personal). I think a better solution, for the future, would be to create a simple module that could update this stubs with more "fresh" JSON data. This way, every few weeks (or whatever strategy we devise), we could run it and end up with fresh JSON data that reflects any potential changes in the data structures of objects being returned by our platform.

My 2 cents.

rapala61:
@aaronschachter 
"have you tried posting some requests to /chatbot locally as sanity test?"
- reported back through my local gambit.

test/stubs/contentful/get-entries-campaign-with-overrides.json
aaronschachter: I think the stub updater that refreshes based on current env variables is a good solution that speaks to the (minor) concern of if we ever changed the start command. Definitely nothing we need to worry about any time soon though.

rapala61: I understand your concern. Let me give you an example, lets say that this updater exists now, and I run this stub updater. The text that you referenced up there would not be affected by it, it would still say: `text START`, because it is not a product of using any variable to set it, it is actually hardcoded in contentful.

It seems like we are hardcoding this variables as part of creating overrides for certain campaign fields in contentful. I believe what we want is to replace that `text START` in contentful for `text {{cmd_reportback}}`. This way, the stub would have the template where `helpers.replacePhoenixCampaignVars` could work its wonderful magic and replace with the appropriate variable (whatever is in the env variable).

There is a possibility that , as we add more JSON mocks, we end up storing a few that do store "rendered" data, which depends on env variables. I think we should run away as much as we can from this pattern, since it introduces what your concern is. You can see how I made a "U" turn and stored the [all-campaign-fields.json](https://github.com/DoSomething/gambit/pull/903/commits/7fea8ed5dec16fbd21bb4e6bfd9a42e44c55a7cd#diff-8ec8b41db0fdf2647ef25e54948245ce) Mock, without the `.rendered` property and just tested that the actual function does populate this property with something, here: [renderAllMessagesForPhoenixCampaign should inject a rendered property to each campaign field](https://github.com/DoSomething/gambit/pull/903/commits/d8d369516fe57ec6164d57e0d73d4d31c23baa06#diff-64a8fd8e22ad28db222c591fde9911eeR383)

aaronschachter: Ahh, yeah - editors aren't adding the {{cmd_reportback}} on overrides. Ideally they should be for if/when we ever change that start Command. On the default campaign, it's set to that (http://ds-mdata-responder-staging.herokuapp.com/v1/campaigns/819 is an example)

lib/contentful.js
```diff
@@ -246,12 +246,11 @@ module.exports.renderAllMessagesForPhoenixCampaign = function (phoenixCampaign)
...
     exports.getAllFieldsForCampaign(phoenixCampaign.id)
-      .then((campaignFields) => { // eslint-disable-line arrow-body-style
-        /**
-         * Use Promise.map to create an array of promises to wait on,
-         * replacing the push->forEach->Promise.all boilerplate
-         */
-        return Promise.map(
+      /**
+       * Use Promise.map to create an array of promises to wait on,
+       * replacing the push->forEach->Promise.all boilerplate
```
aaronschachter: Does this comment make sense on its own (outside of the pull request where we could see the `push` / `Promise.all` it's replacing)?

PR #909, opened 2017-05-23
Check Retry Count before logging incoming messages

### LINKED ISSUE #897: Use Blink Retry Count to fix inaccurate message counts in analytics
_BUG_
-------------------
#### Current Behavior
-When Blink retries a failed request, we're tracking the incoming message for each retry.. suggesting we've received 83 incoming messages that say `'because i like big chickens. keep them toasty my dude'` when it's really only been 1 request that keeps failing.

#### Desired Behavior
- Track each unique incoming message once.

#### Steps to Replicate
- Needs to be a Reportback Submission that keeps failing. In particular, [Signup 3391386](https://www.dosomething.org/signup/3391386) has been failing for every request. Figuring out why is a separate issue.

#### Why This Matters
- If we want incoming message analytics data to be accurate, we should only log an incoming message to analytics during its first Blink attempt. We'd need Blink to pass through a retry count to know when to track the incoming message.

cc @sergii-tkachenko @justkika

aaronschachter:
#### What's this PR do?
* Only tracks an incoming message (and an outgoing message of `broadcast` if incoming message is responding to a broadcast) for the first Blink attempt, fixing the count of incoming messages in Dashbot analytics.
 
* Post locally to `chatbot` and include a `x-blink-retry-count` of 1 or more in the headers. 
    * Verify Dashbot analytics are not sent for the incoming messages
    * Verify Dashbot analytics are sent when the ``x-blink-retry-count` header is removed

#### Relevant tickets
Fixes #897

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -254,7 +254,7 @@ router.use((req, res, next) => {
  */
 router.use((req, res, next) => {
   // If this is a retry, we don't want to track duplicate analytics for this request.
-  const retryCount = Number(req.headers['x-blink-retry-count']);
+  const retryCount = Number(req.get('x-blink-retry-count'));
   if (retryCount && retryCount >= 1) {
     return next();
   }
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -253,6 +253,12 @@ router.use((req, res, next) => {
  * Track incoming message (and outgoing, if this is a reply to a broadcast).
  */
 router.use((req, res, next) => {
+  // If this is a retry, we don't want to track duplicate analytics for this request.
+  const retryCount = Number(req.headers['x-blink-retry-count']);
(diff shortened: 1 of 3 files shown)


app/routes/chatbot.js
```diff
@@ -253,6 +253,12 @@ router.use((req, res, next) => {
  * Track incoming message (and outgoing, if this is a reply to a broadcast).
  */
 router.use((req, res, next) => {
+  // If this is a retry, we don't want to track duplicate analytics for this request.
+  const retryCount = Number(req.headers['x-blink-retry-count']);
```
sergii-tkachenko: It's better to use [`req.get()`](https://expressjs.com/en/api.html#req.get) to access headers, as it performs the case-insensitive search.

aaronschachter:
Good call @sergii-tkachenko -- updated!

PR #910, opened 2017-05-24
Send Error Message and suppress Blink retries on failed Phoenix POST request

### LINKED ISSUE #902: Error Occurred Message Type + Alerts
_FEATURE OVERVIEW_
-------------------
#### User Story
As a member of the O'Doyle Family, I want members who are stuck in Campaign Conversation to receive a message that an error occurred and that we're looking into it. I also want:
*  To be alerted (via Slack or email) that an Error Occurred message was sent (and the DS User / MC Profile ID it was sent to) 
* To let Blink know it shouldn't retry this error message 100 times because the request will fail every time if it's due to data issues (#901, #898, #881)
* To manage the copy for the Error Occurred as a new field on the Contentful Campaign content type, `ErrorOccurredMessage`.

#### Additional Information
We plan to send a custom header in the response to Blink to know whether to retry a response with status code 500, so we'll need a new Blink issue to check for it and retry accordingly once we're sending it through.

TODO:
* [ ] Add a `ErrorOccurredMessage` to the Contentful campaign type, and set a value for the default Campaign
* [ ] Identify scenarios where networking errors like #901, #898 will always return an error and then:
    * [ ] End the chatbot conversation with the Campaign's Error Occurred Message
    * [ ] Trigger an alert somehow and somewhere, like Stathat, Slack, Email, Blink, Twitter, smoke signal, etc 💭  
    * [ ] Include a new request header `x-blink-retry-suppress` with value set to `true` 

#### Why This Matters
Users used to receive a "Whoops, an error occurred message" via Mobile Commons, when our chatbot mData would post to Gambit and get an error back. 

Current behavior is that the user doesn't receive any message back at all, because Blink sends a success response to Mobile Commons indicating the message is queued. When Gambit responds with an error back to Blink, the user doesn't get a response back from us at all (and we wouldn't want them to receive 100 "An error occurred" messages for each Blink retry when the message is destined to fail).

### LINKED ISSUE #907: Text with emoji cause Phoenix API errors
_BUG_
-------------------
#### Current Behavior
- Submitting a caption or why_participated can break Phoenix API with a `PDOException: SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xF0\x9F\xA4\x96' for column 'why_participated' at row 1 in `due to https://github.com/DoSomething/phoenix/issues/6167

#### Desired Behavior
- Either inform end user they can't submit emoji and re-prompt for caption/why 🙅 
- Strip emoji from the input to save successfully

#### Background info
I think this error started because of Mobile Commons switching our SMS provider over Twilio. Previously we'd see emoji come through as question marks and wouldn't run into this.

aaronschachter:
#### What's this PR do?
When a POST Signup or Reportback request fails with status 500:
* Sends a new message type, `error_occurred`, to the end user and places them into Agent View
* Sets the response `x-blink-retry-suppress` to `true` 

Strips emoji from reportback text to fix #907 

Upon staging deploy, text an emoji for a caption and verify that
* Error Occurred Message is sent
* Blink does not retry

Need to update Contentful tests. Opening now to test on staging.

#### Relevant tickets
Fixes #902, #907

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -145,6 +145,17 @@ function endConversationWithError(req, res, error) {
 }
 
 /**
+ * Handles a Phoenix POST error, calling endConversationWithError if we shouldn't retry.
+ */
+function handlePhoenixPostError(req, res, err) {
+  if (err.message.includes('API response is false')) {
+    return endConversationWithError(req, res, err);
+  }
+
+  return helpers.sendErrorResponse(res, err);
+}
+
+/**
  * Check for required config variables.
  * TODO: This MUST be refactored. Why are we checking these env variabes exist on each request?
  */
@@ -447,13 +458,7 @@ router.use((req, res, next) => {
 
       return next();
     })
-    .catch((err) => {
-      if (err.status && err.status === 500) {
-        return endConversationWithError(req, res, err);
-      }
-
-      return helpers.sendErrorResponse(res, err);
-    });
+    .catch(err => handlePhoenixPostError(req, res, err));
 });
 
 /**
@@ -598,13 +603,7 @@ router.post('/', (req, res) => {
 
       return continueConversationWithMessageType(req, res, 'menu_completed');
     })
-    .catch((err) => {
-      if (err.status && err.status === 500) {
-        return endConversationWithError(req, res, err);
-      }
-
-      return helpers.sendErrorResponse(res, err);
-    });
+    .catch(err => handlePhoenixPostError(req, res, err));
 });
 
 module.exports = router;
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -2,6 +2,7 @@
 
 // Third party modules
 const newrelic = require('newrelic');
+const emojiStrip = require('emoji-strip');
 const express = require('express');
 const logger = require('winston');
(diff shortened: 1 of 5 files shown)


aaronschachter:
Needs more testing/investigation -- I've tested on staging by triggering an error with request `4d4a8793-3790-4129-97f6-067dabeb96cd` but doesn't seem to be suppressing retries.

aaronschachter:
Added quick fix for #907 by adding `strip-emoji` -- but having a tough time writing a spy/stub for the `strip-emoji` library in `helpers.test.js`.  The library exports a single function. Also ran into same trouble writing tests for the zendesk branch that behaves similarly. Not having luck getting `sinon.createSubInstance` to work.

aaronschachter:
We're still seeing failed reportback requests for Blink retries after today's deployment because this PR strips them from incoming messages -- but reportback submissions created before today's employment will still contain emoji in the database, that we're still retrying and seeing failures for.

The fix for existing Reportback Submissions that are failing would be to [strip the data before we post it to Phoenix](https://github.com/DoSomething/gambit/blob/4.3.1/app/models/Signup.js#L244)

PR #915, opened 2017-05-26
Submit Zendesk ticket upon incoming Member Support command

### LINKED ISSUE #896: Option to send support requests to Zendesk instead of Agent View
_FEATURE OVERVIEW_
-------------------
#### User Story
As a Member Support staffer, I want to receive and respond to Gambit support requests using Zendesk instead of Agent View.

#### Current Implementation
When a member texts the Member Support Command ("Q"), they are opted into the Chatbot Agent View Opt-in Path. All incoming messages from the User are placed into the Agent View queue.  If the User wants to continue their Gambit Campaign conversation, they must text back the Signup Keyword.

#### TODO
* [ ] When a member texts Member Support Command, their message is posted to the Zendesk API as a new ticket.
* [ ] After the member sends their support question, Gambit responds with a new `member_support_confirmation` message, informing the User their message will be answered by a DoSomething staffer from a different phone number (e.g. _"Thanks for the question, we've created a support ticket and we'll try to text back an answer within 24 hrs. Text BOOKBOT to continue Two Books, or QUESTION if you need to add more info to your ticket."_) 

    * All subsequent incoming messages are answered by Gambit: If the user didn't send a valid Start or Member Support command, Gambit will reply with the a "Sorry, I didn't get that" + the Member Support Confirmation message

* [ ] We need the ability to switch this feature off completely via config variable, to return sending support requests to Agent View 

* [ ] BONUS Filtering: If the support ticket contains any explicit language, don't create the ticket but respond a prompt to reword their question appropriately (#895)

cc @jamjensen @justkika @classicfreddie

aaronschachter:
#### What's this PR do?
Creates a Zendesk ticket when a user sends an incoming message which is a Member Support Command, when new config variable `MEMBER_SUPPORT_PROVIDER` is set to `zendesk`.  This triggers a text from our Zendesk text number, asking the user to text back their question.

Under the hood, here's what's going on:

* Gambit creates a new Zendesk ticket with a `gambit` tag by posting to the Zendesk API
* I've set up a trigger in Zendesk called "Begin SMS Support" that will send a text confirmation to the requester any time a ticket with `gambit` is created. The text confirmation contains the Member Support intro. It also automatically closes the newly created `gambit` ticket.
* When a member responds to the text confirmation, a new ticket is created in Zendesk. When agents respond to it in Zendesk, their response is sent as SMS via Zendesk text.

screen shot 2017-05-26 at 2 09 59 pm

We may never end up turning this on for production or even merging into `develop`, this branch is opened to test on Thor.

Locally:
 * Pull down this branch and set new config variables 
 * POST Member Support command as `args` to `/chatbot`

Staging:
* Begin/continue a Thor Campaign Conversation 
* Text a Member Support command 

**Testing**
I've been struggling trying to stub a constructor for the Zendesk library in the `zendesk.test.js`. I similarly had a tough time trying to add tests to mock calling `emoji-strip` in #910.

**Zendesk API**
My ideal workflow would be that upon texting a member support command, a member sends their question, and then we create a Zendesk ticket and respond with a "We'll get back to you as soon as we can" -- exactly like how the Member Support Message works currently except creating the Zendesk ticket instead of placing the member into Agent View.

The reason that won't fly is that we can't respond to that newly created ticket using Zendesk Text. We can only respond via SMS when the ticket is created via SMS, which is why we use 2 (the first being the throwaway ticket to create a trigger with). 

It's got me wondering if the [Front API](http://dev.frontapp.com/) would enable implementing my ideal workflow.

#### Relevant tickets
Fixes #896 

#### Checklist
- [x] Tested on staging.


aaronschachter:
No go on Zendesk for the workarounds documented in the PR description and #896, refs https://www.pivotaltracker.com/n/projects/1979505/stories/145639137

PR #917, opened 2017-06-05
Chatbot testing

rapala61:
#### What's this PR do?
I have started to refactor middleware in the chatbot route. It adds test coverage for these middleware. It also has some refactoring.

- 👀 
- Run `npm run all-tests`
- Use postman to signup and reportback to a bot campaign. It should work with no problemo.

A lot of movement in the code.

#### Checklist
- [x ] Tested on localhost by signing up and reporting back to SEEDBOT and LUCKYBOT.


app/routes/chatbot.js
```diff
@@ -13,13 +12,24 @@ const phoenix = require('../../lib/phoenix');
 const stathat = require('../../lib/stathat');
 const ClosedCampaignError = require('../exceptions/ClosedCampaignError');
 
+// config modules
+const requiredParamsConf = require('../../config/middleware/chatbot/required-params');
+const requestContextConf = require('../../config/middleware/chatbot/request-context');
```
aaronschachter: What do you think about renaming this to something like `request-input-params`? `Context` feels so generic, but I guess it brings up the question of maybe it's the place to store loading a Campaign per an incoming Broadcast ID too.

app/routes/chatbot.js
```diff
@@ -159,134 +169,21 @@ function handlePhoenixPostError(req, res, err) {
...
-        return helpers.sendTimeoutResponse(res);
-      }
-      req.user = user; // eslint-disable-line no-param-reassign
-      newrelic.addCustomParameters({ userId: user._id });
-
-      return next();
-    })
-   .catch(err => helpers.sendErrorResponse(res, err));
-});
+router.use(RegisterNewUserIfNotFound());
```
aaronschachter: Similar question here -- I'm a little hesitant to refer to Gambit making register User requests... vs create User requests, which maps exactly to the post we're making to Northstar.

aaronschachter:
```
get-user-by-phone.js
 map-request-params.js
register-new-user.js
request-context.js
required-params.js
timeouts.js
```
Another minor naming thing, but I'm considering whether some naming changes here might make its easier to scan what's going on each, e.g renames to `user-get.js, `user-create.js`.  Two of the files use `request` in the name...one could argue each of these middleware files uses the incoming request too though, no?

app/routes/chatbot.js
rapala61: This is exactly the place where I did some soul searching about the name. I agree the naming is generic. I was thinking something like `request-conversation-context`? The `request-input-params` is better than what is there now, but still generic. The `requiredParamsConf ` middleware takes care of making sure required params are there, for now `phone` is the only non-negotiable param. We then look for one of 3: `args`, `keyword`, or `mms_image_url`. It seems to me that, the "context" or this request/conversation is defined by which of these params is included in the request, thats why i named it that way.

rapala61:
@aaronschachter Yeah, I can do a better job with the naming of some middleware. I'll update and revise them.

app/routes/chatbot.js
aaronschachter: I definitely like where Conversation is headed. This also handles generally whatever the incoming message is, whether it be only a keyword, or only an image (or combos).. spitballing ideas here:
* `incoming-message-params`
* `user-incoming-message` (could group nicely with `user-get`, `user-create`)

PR #918, opened 2017-06-05
start troubleshooting forbidden signup post issue

rapala61:
#### What's this PR do?
It's just a placeholder PR to create a test heroku app to the troubleshoot another classic "It works in my machine" episode. 
- The error:
    - `Signup.post error: Forbidden`
- Does it happen when signing up in localhost?
    - Nope.
- Does it happen when signing up through Thor?
    - Oh Yes.
- Why is this happening:
    - Not sure yet, lets try to isolate the problem through this test app.

- Not yet

### DIFF
--- a/lib/middleware/map-request-params.js
+++ b/lib/middleware/map-request-params.js
@@ -1,5 +1,6 @@
 'use strict';
 
+// start troubleshooting here
 const emojiStrip = require('emoji-strip');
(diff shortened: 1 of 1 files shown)


rapala61:
Context for the problem and solution are in this PR: https://github.com/DoSomething/phoenix/pull/7403

PR #924, opened 2017-06-14
More middleware refactoring

rapala61:
#### What's this PR do?
More code moved out of the chatbot route

- 👀 
- Run `npm run all-tests`

#### Checklist
- [x] Tested on localhost.

### DIFF
--- a/app/routes/chatbot.js
+++ b/app/routes/chatbot.js
@@ -7,10 +7,6 @@ const logger = require('winston');
 
 // Application modules
 const helpers = require('../../lib/helpers');
-const contentful = require('../../lib/contentful');
-const phoenix = require('../../lib/phoenix');
-const stathat = require('../../lib/stathat');
-const ClosedCampaignError = require('../exceptions/ClosedCampaignError');
 
 // config modules
 const requiredParamsConf = require('../../config/middleware/chatbot/required-params');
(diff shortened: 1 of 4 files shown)


lib/middleware/map-request-params.js
```diff
@@ -70,13 +70,13 @@ function parseKeyword(req, res) {
...
 function parseCampaignId(req) {
+  let campaign = {};
   if (req.broadcast_id && req.broadcastContentfulObject) {
-    const broadcastCampaign = getCampaignFromContentfulObject(req.broadcastContentfulObject);
-    req.campaignId = broadcastCampaign.campaignId;
+    campaign = getCampaignFromContentfulObject(req.broadcastContentfulObject);
   } else if (req.keyword && req.keywordContentfulObject) {
-    const keywordCampaign = getCampaignFromContentfulObject(req.keywordContentfulObject);
-    req.campaignId = keywordCampaign.campaignId;
+    campaign = getCampaignFromContentfulObject(req.keywordContentfulObject);
```
aaronschachter: This function name feels a little vague, since we have different Phoenix and Contentful Campaigns objects. What do you think about renaming to something like `getContentfulCampaignFromObject`. `getContentfulCampaignFromContentfulObject` is awkward for sure (but I can see the advantage of keeping the `ContentfulObject` naming since you've defined it that way in other places).

app/routes/chatbot.js
```diff
@@ -24,143 +20,17 @@ const mapRequestParamsMiddleware = require('../../lib/middleware/map-request-par
 const lookUpUserMiddleware = require('../../lib/middleware/user-get');
 const createNewUserIfNotFoundMiddleware = require('../../lib/middleware/user-create');
 const processBroadcastConversationMiddleware = require('../../lib/middleware/broadcast');
+const getPhoenixCampaignMiddleware = require('../../lib/middleware/campaign-get');
```
aaronschachter: Similar thought, would it make sense to specifically spell this out as `phoenix-campaign-get` in the filename since getting a Campaign could potentially refer to querying Contentful.

rapala61:
@aaronschachter Thanks! I'll address the name changing in the next PR.

PR #926, opened 2017-06-28
Integration testing

### LINKED ISSUE #927: Home route TypeError
_BUG_
-------------------
#### Current Behavior
- GET '/'
TypeError: stathat.postStat is not a function
    at router.get (/app/app/routes/home.js:9:11)

rapala61:
#### What's this PR do?
This is a biggie.
This PR introduces our first integration tests as a test drive for the more important/complex tests.
Also, it refactors the way Gambit replies back to the use. It uses a different pattern.

- 👀 
- Test all actions for which the user would get back a response: signup, signup for closed campaign, reportback flow, wrong inputs for qty, and why, no photo, broadcast y/n, etc.

#### Checklist
- [x] Tested on localhost.

#### Relevant Issues
- Fixes #927

### DIFF
--- a/lib/conversation/replies.js
+++ b/lib/conversation/replies.js
@@ -3,6 +3,7 @@
 /* eslint-disable no-param-reassign */
 
 const logger = require('winston');
+const newrelic = require('newrelic');
 
 const Command = require('./command');
 const actions = require('./actions');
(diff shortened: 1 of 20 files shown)


lib/middleware/user-support-conversation.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const helpers = require('../helpers');
+const ReplyDispatcher = require('../conversation/reply-dispatcher');
+const replies = require('../conversation/replies');
+
+module.exports = function processBroadcast() {
```
aaronschachter: `processSupportConversation()` ?

test/integration/routes/status.test.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+// libs
+require('dotenv').config();
+const test = require('ava');
+const chai = require('chai');
+// const sinon = require('sinon');
+const sinonChai = require('sinon-chai');
+// const httpMocks = require('node-mocks-http');
```
aaronschachter: Any reason to keep these commented `const`statements in?

lib/conversation/replies.js
```diff
@@ -0,0 +1,148 @@
...
+        args.req.user.postDashbotOutgoing(args.text);
+        stathat.postStat(`campaignbot:${args.text}`);
+        BotRequest.log(args.req, 'campaignbot', type, args.text);
+        return helpers.sendResponse(args.res, 200, args.text);
+      })
+      .catch(err => helpers.sendErrorResponse(args.res, err));
+}
+
+function renderContinueConversationMsg(fn) {
+  return (type, args) => helpers.renderMessageForMessageType(args.req, type)
```
aaronschachter: What do you think about renaming `type` to `messageType`? I think it would help with readability. 

`messageTemplate` could also be a contender... I defined this property as [`template` in the Slothie `Message` model](https://github.com/DoSomething/slothie/blob/264a3a8df76bae44240e039eb0884128de891538/app/models/Message.js#L18), inspired how Dashbot lets you pass an optional [Template ID](https://www.dashbot.io/sdk/template) in their message analytics.

lib/conversation/replies.js
```diff
@@ -0,0 +1,148 @@
...
+  return (args) => {
+    fn(args);
+    return helpers.sendResponse(args.res, 200, args.text);
+  };
+}
+
+function renderEndConversationMsg(fn) {
+  return (type, args) => helpers.renderMessageForMessageType(args.req, type)
+      .then((replyMessage) => {
+        args.text = replyMessage;
```
aaronschachter: Similar comment / concern about readability here --  What do you think about renaming this to `args.replyText`?  Because `req` and `res` get both passed as `args`, the general `text` property was throwing me off a little bit. Since we're passing both `req` and `res`  - is there a reason for defining `text` property in `args` vs adding it to the `req` ?

aaronschachter:
This branch seems to fix #927.

lib/conversation/replies.js
rapala61: I'm sending the `req`, `res`, and `replyText` inside an `args` object so I don't have to bother with the variadic nature of the replies. The point here is that each reply gets all the "ingredients" necessary to return a new function (a command) that the dispatcher executes.

I could have also approached this by sending arguments as needed:
- `replyDispatcher.execute(replies.legacyBugErrorOcurred(req, res, err))`
Instead of:
- `replyDispatcher.execute(replies.legacyBugErrorOcurred({ req, res, error: err }))`
Then define each reply and decorators with the needed parameters, etc. Which I think It would make the code more complex:
    ```
    module.exports.legacyBugErrorOcurred = function legacyBugErrorOcurred(req, res, error) {
      return new Command(
        renderEndConversationWithErrorMsg(actions.endConversation),
         ['error_occurred', req, res, error]);
    };
```

rapala61:
@aaronschachter it does fix #927. Found the bug once I created the integration test for that route.

rapala61:
@aaronschachter made some changes you suggested around naming: `replyText`, and `messageTemplate`.

lib/conversation/replies.js
aaronschachter: I'm on board with a single `args` object -- it was more a question of potentially defining `replyText` on our `req` object (like we do for our `req.user`, `req.campaign` etc) that we pass to `args`. I'm not a strong advocate of one or the other but was just curious about the choice to define as `args` property vs adding to `req`. Regardless of which object it's defined in--  what do you think about renaming it from `text` to `replyText`

PR #930, opened 2017-07-07
fix to get menu_completed showing up in stathat again

rapala61:
#### What's this PR do?
Fixes stathat logging when sending replies to users.

- 👀 

#### Checklist
- [ ] Tested on staging.
- [ ] Tested locally.

### DIFF
--- a/lib/conversation/replies.js
+++ b/lib/conversation/replies.js
@@ -99,7 +99,7 @@ function renderContinueConversationMsg(fn) {
           .then(() => {
             logger.verbose(`saved user:${args.req.user._id} current_campaign:${args.req.user.current_campaign}`);
             fn(args);
-            args.req.user.postDashbotOutgoing(args.replyText);
+            args.req.user.postDashbotOutgoing(messageTemplate);
             stathat.postStat(`campaignbot:${messageTemplate}`);
             BotRequest.log(args.req, 'campaignbot', messageTemplate, args.replyText);
             return helpers.sendResponse(args.res, 200, args.replyText);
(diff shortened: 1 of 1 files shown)


lib/conversation/replies.js
```diff
@@ -100,7 +100,7 @@ function renderContinueConversationMsg(fn) {
             logger.verbose(`saved user:${args.req.user._id} current_campaign:${args.req.user.current_campaign}`);
             fn(args);
             args.req.user.postDashbotOutgoing(args.replyText);
```
aaronschachter: The Dashbot analytics should be tracking the messageTemplate and not the actual rendered message:
screen shot 2017-07-07 at 8 54 04 am

aaronschachter:
If we're sticking with Stathat (which.. is questionable if anyone ever looks at Stathat anymore... cc @jamjensen) we need to manually delete all the Stathat stats we've been logging with message text since that deploy:
https://www.stathat.com/v/stats/EIYS
screen shot 2017-07-07 at 8 57 22 am

rapala61:
@aaronschachter made the update for both loggers.

aaronschachter:
Won't we still see the same behavior for the `renderEndConversationMsg` functions?

rapala61:
Yes we would. Changing those now.

rapala61:
@aaronschachter what about now?

jamjensen:
@aaronschachter Sorry just seeing this. I have not used the stathat reports in sometime. Still interested in seeing the breakdown by message type (ask_quantity, ask_photo, reason_why) so that we're able to look into potential drop-off. If another service does a better job then I'm totally game.

PR #932, opened 2017-07-12
User.last_outbound_template

aaronschachter:
#### What's this PR do?
Adds a `last_outbound_template` property to our User model, and updates it when sending a reply back.

Post to chatbot and verify your User's `last_outbound_template` is updating accordingly.

This is in prep for adding Rivescript support into Gambit, when we'll want to reply differently based on whether the last message a User received was from Rivescript vs a Gambit message template.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/conversation/replies.js
+++ b/lib/conversation/replies.js
@@ -43,11 +43,15 @@ function renderEndConversationMsg(fn) {
   return (messageTemplate, args) => helpers.renderMessageForMessageType(args.req, messageTemplate)
       .then((replyMessage) => {
         args.replyText = replyMessage;
-        fn(args);
-        args.req.user.postDashbotOutgoing(messageTemplate);
-        stathat.postStat(`campaignbot:${messageTemplate}`);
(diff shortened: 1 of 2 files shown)


lib/conversation/replies.js
```diff
@@ -95,7 +104,7 @@ function renderContinueConversationMsg(fn) {
         // Store current campaign to continue conversation in subsequent messages.
         // TODO: Should this be handled more eloquently?
         args.req.user.current_campaign = args.req.campaignId;
```
aaronschachter: I was debating whether it made sense to pass the `req.campaignId` as a second optional parameter to `setLastOutboundTemplate` to save it there... and whether it would add/hurt eloquence...

rapala61: Hmm, what about if we create an schema method to save the current campaign, just like how saving the last outbound template works. Then we could use a Promise.all to proceed with the reply? Not a blocker for merging.

aaronschachter:
Good idea! Will do in a future PR.

PR #934, opened 2017-07-17
User.updateConversation

aaronschachter:
#### What's this PR do?
* Refactors `User.setOutboundTemplate(messageTemplate)` as `User.updateConversation(messageTemplate, campaignId)`, setting the User:
    * Last Outbound Template -- The last message template the User received
    * Current Campaign
    * Signup Status -- a new property, set to `'doing'` by default if a `campaignId` is passed. As we look to handle more responses than just "Yes" an Ask Signup Message, we'll want to save values like `'ask_signup'` or `'declined_signup'` here to determine how to reply.

* DRY - Refactor the `signups` router to use `lib/conversations/replies` for External Signup Menu Message delivery.  This route was missed in #932 

* Rename  `helpers.renderMessageForMessageType` as `helpers.renderMessageTemplate` for better readability, as ideally all references to `messageType` should eventually get renamed to `messageTemplate`. 

* Full chatbot conversation to test signing up, changing current campaigns, and conversation enders (e.g. send a Closed Campaign keyword)
* Test a POST `/signups` request with a Signup for your User to verify the External Signup Message is sent (and your Campaign and Last Outbound Template are updated)

Signup Broadcast functionality currently ends the conversation if the User says a non-YES to an incoming Broadcast.

Potential game plan for our chatbot to continue the conversation instead is:
* Set the User's Current Campaign to the broadcast campaign
* If they responded with a NO, send a new Campaign `askSignupDeclinedMessage` and set their Signup Status to `signup_declined`.
* If responded with neither YES or NO, send a new Campaign `invalidAskSignupResponseMessage` to the User, and set their Signup Status to `'ask_signup'`.

This sets the Current Campaign to the Broadcast Campaign so we can eventually listen for any Campaign specific questions, and still allow the User to say YES to the Signup after we've answered the question -- the MVP of this approach is [here in Slothie.](https://github.com/DoSomething/slothie/blob/master/app/routes/chatbot.js)

#### Relevant tickets
Refs https://github.com/DoSomething/gambit/pull/932#discussion_r127262532

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/models/BotRequest.js
+++ b/app/models/BotRequest.js
@@ -57,7 +57,7 @@ botRequestSchema.statics.log = function (req, botType, msgType, msg) {
     data.user_id = req.user._id;
   }
   if (req.campaign) {
-    data.campaign_id = req.campaign._id;
+    data.campaign_id = req.campaign.id;
   }
   data.bot_type = botType;
   data.bot_response_type = msgType;
@@ -69,7 +69,7 @@ botRequestSchema.statics.log = function (req, botType, msgType, msg) {
 
       return doc;
     })
-    .catch(err => logger.err(`BotRequest error:${err.message}`));
+    .catch(err => logger.error(`BotRequest error:${err.message}`));
 };
 
 module.exports = mongoose.model('bot_requests', botRequestSchema);
--- a/app/models/User.js
+++ b/app/models/User.js
@@ -28,6 +28,12 @@ const userSchema = new mongoose.Schema({
   first_name: String,
   // Campaign the user is currently participating in via chatbot.
   current_campaign: Number,
+  signup_status: {
+    type: String,
+    // Eventually we'll be adding other values here like 'prompt, 'declined' to build out handling
+    // No or Invalid responses to a Ask Signup Message Template message (broadcasts, campaign menu).
+    enum: ['doing'],
+  },
   last_outbound_template: String,
 
 });
@@ -168,6 +174,7 @@ userSchema.methods.updateConversation = function (outboundTemplate, campaignId) 
 
   if (campaignId && campaignId !== this.current_campaign) {
     this.current_campaign = campaignId;
+    this.signup_status = 'doing';
   }
 
   return this.save();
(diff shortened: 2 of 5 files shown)


app/models/BotRequest.js
```diff
@@ -57,7 +57,7 @@ botRequestSchema.statics.log = function (req, botType, msgType, msg) {
     data.user_id = req.user._id;
   }
   if (req.campaign) {
-    data.campaign_id = req.campaign._id;
+    data.campaign_id = req.campaign.id;
```
aaronschachter: There used to be a `Campaign` model that kept a cache of the Phoenix Campaigns. We ended up removing it to eliminate complexity, although I've been daydreaming more and more about a Gambit Campaigns microservce to power the GET `/campaigns` and `/campaigns/:id` endpoints that the Gambit chatbot endpoint can query to get a rendered reply message for a given Campaign and template.

aaronschachter:
Write up on finding a Signup ID to test with -- https://github.com/DoSomething/gambit/wiki/Tests#external-signups

app/models/User.js
```diff
@@ -159,8 +165,17 @@ userSchema.methods.postDashbotOutgoing = function (gambitMessageType) {
   this.postDashbot('outgoing', gambitMessageType);
 };
 
-userSchema.methods.setLastOutboundTemplate = function (messageTemplate) {
-  this.last_outbound_template = messageTemplate;
+/**
+ * @param {string} messageTemplate
+ * @param {number} campaignId
+ */
+userSchema.methods.updateConversation = function (outboundTemplate, campaignId) {
```
rapala61: Not a blocker for merging, but I like what you did here: https://github.com/DoSomething/slothie/blob/master/app/models/User.js#L90-L120. (splitting up functions for statuses) Maybe we can bring that same logic here?

aaronschachter: Will do in a next PR. Thanks for the review, and testing!

PR #944, opened 2017-08-12
Receive Message endpoint

### LINKED ISSUE #939: Receive message endpoint
_FEATURE OVERVIEW_
-------------------
#### User Story
As an ODev, I want to post an inbound [Gambit Conversations](https://github.com/dosomething/gambit-conversations) message to a new [Gambit Campaigns](https://github.com/dosomething/gambit) `/receive-message` endpoint -- a scaled-down version of the `chatbot` endpoint, which creates or updates the User's Signup for the given Campaign ID, and responds with the reply text and template to send back to the User.

#### Additional Information

Input parameters:
* `phone`
* `campaignId`
* `text`
* `attachments`
* `keyword`
* `broadcastId`

External `/receive-message` requests:
* Find or create Northstar User by `phone`
* Find or create Phoenix Signup for User for `campaignId`
* Post Phoenix Reportback for Signup if User has completed conversation
* Get rendered reply message from Phoenix/Contentful

What we're NOT doing:
* Getting Campaign by `req.body.keyword`, `req.body.broadcast`, or `User.current_campaign`
* Posting the reply message to Mobile Commons

#### Why This Matters

The Conversations API will be the source of Conversation State for a given User in the near future -- it will no longer be the responsibility of the Gambit Campaigns API to determine which Campaign a User is completing/signing up for.

The Conversations API is in development and only open to staff. It's currently posting to the Gambit Campaigns `/chatbot` endpoint to get reply messages by sending a [invalid mobile number (our Conversation Mongo ID) as a `phone` parameter](https://github.com/DoSomething/gambit-conversations/blob/0.2.0/lib/middleware/receive-message/send-gambit-reply.js#L17). Because it's an invalid phone number, the post to Mobile Commons fails (and Conversations API is the only app sending messages back to the end user)

aaronschachter:
#### What's this PR do?
Adds a **POST** `/receive-message` endpoint that acts like **POST** `/chatbot`, with a few differences:
* It expects a `campaignId` parameter to determine which Campaign to create a Signup for (or to save Signup data against, e.g. quantity, photo)
* It doesn't post to Mobile Commons to send the Gambit reply back to the User-- sending the reply will be the responsibility of the Conversations resource that posts here  https://github.com/DoSomething/gambit-conversations/issues/59

Post locally to `/receive-message` to verify Signup + Campaign Completion steps.

```
curl -X "POST" "http://localhost:5000/v1/receive-message" \
     -H "x-gambit-api-key: [apiKey]" \
     -H "Content-Type: application/x-www-form-urlencoded; charset=utf-8" \
     --data-urlencode "phone=[yourPhoneNumber]" \
     --data-urlencode "text=3" \
     --data-urlencode "campaignId=2900"
 ```

This endpoint will deprecate the `/chatbot` endpoint (and `User` model) once Conversations launches - our [Conversation model will store the Campaign](https://github.com/DoSomething/gambit-conversations/blob/1158a34dc98da6ddee347e6fbcd475bea29b7c69/app/models/Conversation.js#L21) a User is currently in a conversation for.

#### Relevant tickets
Fixes #939 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/config/middleware/receive-message/map-request-params.js
+++ b/config/middleware/receive-message/map-request-params.js
@@ -1,0 +1,15 @@
+'use strict';
+
+const configVars = {};
+
+configVars.client = 'gambit-conversations';
+configVars.containerProperty = 'body';
+configVars.emojiStripParam = 'text';
+configVars.lowercaseParam = 'text';
+configVars.paramsMap = {
+  keyword: 'keyword',
+  broadcastId: 'broadcast_id',
+  text: 'incoming_message',
+  mediaUrl: 'incoming_image_url',
+};
+module.exports = configVars;
--- a/lib/middleware/receive-message.js
+++ b/lib/middleware/receive-message.js
@@ -1,0 +1,9 @@
+'use strict';
+
+module.exports = function receiveMessage() {
+  return (req, res, next) => {
+    req.campaignId = req.body.campaignId;
+
+    return next();
+  };
+};
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -3,6 +3,7 @@
 // Routes
 const campaignsRoute = require('./campaigns');
 const chatbotRoute = require('./chatbot');
+const receiveMessageRoute = require('./receive-message');
 const signupsRoute = require('./signups');
 const statusRoute = require('./status');
 const homeRoute = require('./home');
@@ -30,5 +31,6 @@ module.exports = function init(app) {
   app.use('/v1/status', statusRoute);
   app.use('/v1/campaigns', campaignsRoute);
   app.use('/v1/chatbot', chatbotRoute);
+  app.use('/v1/receive-message', receiveMessageRoute);
   app.use('/v1/signups', signupsRoute);
 };
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -29,8 +29,16 @@ module.exports = function init(app) {
   regGlobalMiddleware(app);
   app.get('/', homeRoute);
   app.use('/v1/status', statusRoute);
+
+  // Provides list of Campaigns currently running on Gambit.
   app.use('/v1/campaigns', campaignsRoute);
+
+  // Receives inbound message from Mobile Commons mData.
   app.use('/v1/chatbot', chatbotRoute);
+
+  // Receives inbound message from Gambit Conversations service.
   app.use('/v1/receive-message', receiveMessageRoute);
+
+  // Sends external Signup message to a User for a Campaign.
   app.use('/v1/signups', signupsRoute);
 };
--- a/app/routes/receive-message.js
+++ b/app/routes/receive-message.js
@@ -1,0 +1,95 @@
+'use strict';
+
+const express = require('express');
+
+// configs
+const requiredParamsConf = require('../../config/middleware/receive-message/required-params');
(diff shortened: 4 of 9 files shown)

### CODE REVIEW

File: config/middleware/receive-message/map-request-params.js [hunk 1]
```diff
@@ -0,0 +1,15 @@
...
+
+configVars.client = 'gambit-conversations';
+configVars.containerProperty = 'body';
+configVars.emojiStripParam = 'text';
+configVars.lowercaseParam = 'text';
+configVars.paramsMap = {
+  keyword: 'keyword',
+  broadcastId: 'broadcast_id',
+  text: 'incoming_message',
+  mediaUrl: 'incoming_image_url',
```
aaronschachter: I tried adding `campaignId: 'campaignId',` here to avoid creating the [`receive-message` middleware](https://github.com/DoSomething/gambit/blob/703fa6d59bde8db4f8ab47247c0e4e7e64ef4804/config/middleware/receive-message/map-request-params.js) -- but the `req.campaignId` wouldn't be defined in the [User Get middleware.](https://github.com/DoSomething/gambit/blob/703fa6d59bde8db4f8ab47247c0e4e7e64ef4804/lib/middleware/user-get.js#L13)

lib/middleware/receive-message.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+module.exports = function receiveMessage() {
+  return (req, res, next) => {
+    req.campaignId = req.body.campaignId;
```
aaronschachter: Refs https://github.com/DoSomething/gambit/pull/944/files#r132808228

config/middleware/receive-message/map-request-params.js
rapala61: I tried this locally and removed the `receive-message` middleware and it seems to work.

aaronschachter: Passing a `campaignId` here does return a rendered reply message -- but changing the `campaignId` parameter value isn't changing the Signup / Campaign accordingly, [it still loads the Campaign from the User.current_campaign. ](https://github.com/DoSomething/gambit/blob/703fa6d59bde8db4f8ab47247c0e4e7e64ef4804/lib/middleware/user-get.js#L13)

rapala61: Maybe I'm not following the issue. By adding the `campaignId` in the `map-request-params` config, the `campaignId` being sent in the POST request `body` gets parsed out and injected into the request object, like `req.campaignId`, which is essentially the same thing the [receive-message](https://github.com/DoSomething/gambit/blob/703fa6d59bde8db4f8ab47247c0e4e7e64ef4804/lib/middleware/receive-message.js) middleware is doing, which is why I mentioned the middleware is not required.

aaronschachter: This was the culprit: https://github.com/DoSomething/gambit/blob/4.6.0/lib/middleware/map-request-params.js#L78 -- we don't need to call `parseCampaignId` from the `receive-message` route.

https://github.com/DoSomething/gambit/pull/945

rapala61: Great catch! I see the problem now.

PR #947, opened 2017-08-17
Hotfix/remove check for mobilecommons in actions

rapala61:
#### What's this PR do?
Fixes mobile commons missing profile bug by removing a check for `client == 'mobilecommons'` in order to post a profile update. Posts to `signups` route are not being processes correctly due to this check. We have to revisit.

- 👀


rapala61:
The ugly release commits included here are a side effect of the strategy we have been using in Gambit Campaigns to this date.

PR #957, opened 2017-09-05
Return data.templates in GET campaigns/:id 

aaronschachter:
#### What's this PR do?
* Renames the `data.messages` property as `data.templates`
* Removes the `Message` suffix from the values returned, to improve data readability 

* Verify `GET /campaigns/:id` returns `data.templates` without a `Message` suffix
* Sanity check that `POST /chatbot` conversations work as expected

Not renaming the single `renderMessage` functions yet. Ideally we should avoid querying Contentful for each request and re-introduce caching Campaigns to this app to improve performance and fail-proofing.

This will break 2 things that will be quick to change, with PR's coming soon:
* [ ] Gambit Conversations sync process
* [ ] Gambit Slackbot "View messages" response

#### Relevant tickets
* Fixes https://github.com/DoSomething/gambit-conversations/pull/99#issuecomment-327223167
* https://github.com/DoSomething/gambit-conversations/issues/49

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -249,8 +249,8 @@ module.exports.getAllTemplatesForCampaignId = function getAllTemplatesForCampaig
 /**
  * Returns all rendered messages for given phoenixCampaign, using defaults when no override defined.
  */
-module.exports.renderAllMessagesForPhoenixCampaign = function (phoenixCampaign) {
-  logger.debug(`renderAllMessagesForPhoenixCampaign:${phoenixCampaign.id}`);
+module.exports.renderAllTemplatesForPhoenixCampaign = function (phoenixCampaign) {
+  logger.debug(`renderAllTemplatesForPhoenixCampaign:${phoenixCampaign.id}`);
   return new Promise((resolve, reject) =>
     // get all fields, overrides and default
     exports.getAllTemplatesForCampaignId(phoenixCampaign.id)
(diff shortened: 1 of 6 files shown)


lib/contentful.js
```diff
@@ -208,31 +208,39 @@ module.exports.renderMessageForPhoenixCampaign = function (phoenixCampaign, msgT
...
         return exports.fetchCampaign(defaultCampaignId);
       })
       .then((defaultContentfulCampaign) => {
         campaignContent.defaults = defaultContentfulCampaign.fields;
         campaignFieldNames.forEach((key) => {
           const fieldName = config.campaignFields[key];
-          campaignFields[fieldName] = {};
+          // This is a hack to avoid migrating our Contentful data in order to rename our fields.
+          // We want to return template values as 'askQuantity', not 'askQuantityMessage'.
+          const templateName = fieldName.split('Message')[0];
```
aaronschachter: I'm going to move this into a `helpers` function and add tests. We should also return the same value in our in the `reply.template` response of `POST /receive-message`:
```
{
  "success": {
    "code": 200,
    "message": "@dev Got it! Now text back a caption for your photo (think Instagram)! Keep it short & sweet, under 60 characters please.",
    "template": "askCaptionMessage"
  }
}
```

aaronschachter: I'll add these in a new PR

Issue #959, opened 2017-09-14
Refactor Receive Message to accept userId parameter


aaronschachter:
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to:
- Accept a `userId` parameter in a `POST /receive-message` request, and use it to find or create a User model, deprecating the need to query Northstar for a User
- Deprecate the `phone` parameter

#### Why This Matters
Each request to the Gambit Conversations `POST /receive-message` will require looking up the Northstar User to determine their subscription status per platform (if they've opted out with status `'unsubscribed'` and are messaging us again, we want to set their status back to `'active'`).  The request we make on the Gambit Campaigns side is extraneous, as it's for the sake of finding the Northstar User ID for a given phone number.

#### Current Workaround w/o Feature (if applicable)
We could keep things as is, but it'd be cleanest to avoid the extra lookup.

PR #962, opened 2017-09-20
Set req.userId

aaronschachter:
#### What's this PR do?

* Sets a `req.userId` in the User Get and User Create middleware

* Modifies `Signup.lookupCurrent` and `Signup.post` to expect a `userId` string instead of a `user` object.
    * This is prep work for #959 - if we pass a `userId` parameter to `POST /receive-message` and set `req.userId`, we can remove the User Get / User Create middleware from Receive Message

* Modifies `Signup.lookupCurrent` and `Signup.post` to expect a `campaignId` number instead of a `campaign` object, to keep consistent with the `userId` change

* Set phone number in Consolebot to a number that doesn't exist for a User (I tend to always use a timestamp, like 201709201232)
* Post a signup , e.g. send `keyword:dunkbot`, to verify a Signup is created
* Continue conversation to verify Signup lookup is working as expected

When we go live with Conversations, we can completely deprecate our `User` model and relevant middleware 🔥 

#### Relevant tickets
#959 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/middleware/signup-get.js
+++ b/lib/middleware/signup-get.js
@@ -4,8 +4,8 @@ const helpers = require('../helpers');
 const Signup = require('../../app/models/Signup.js');
 
 module.exports = function getSignup() {
-  return (req, res, next) => {
-    Signup.lookupCurrent(req.userId, req.campaignId)
+  return (req, res, next) => { // eslint-disable-line arrow-body-style
+    return Signup.lookupCurrent(req.userId, req.campaignId)
       .then((signup) => {
         helpers.handleTimeout(req, res);
 
--- a/lib/middleware/signup-get.js
+++ b/lib/middleware/signup-get.js
@@ -5,9 +5,8 @@ const Signup = require('../../app/models/Signup.js');
 
 module.exports = function getSignup() {
   return (req, res, next) => {
-    const user = req.user;
     const campaign = req.campaign;
-    return Signup.lookupCurrent(user, campaign)
+    return Signup.lookupCurrent(req.userId, campaign)
       .then((signup) => {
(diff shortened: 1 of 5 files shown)


lib/middleware/signup-get.js
```diff
@@ -4,15 +4,13 @@ const helpers = require('../helpers');
 const Signup = require('../../app/models/Signup.js');
 
 module.exports = function getSignup() {
-  return (req, res, next) => {
-    const user = req.user;
-    const campaign = req.campaign;
-    return Signup.lookupCurrent(user, campaign)
+  return (req, res, next) => { // eslint-disable-line arrow-body-style
+    return Signup.lookupCurrent(req.userId, req.campaignId)
```
aaronschachter: Without this `return`, the test fails:
```
  lib › middleware › signup-get › getSignup should call sendErrorResponse when an error occurs
  /pipeline/source/test/lib/middleware/signup-get.test.js:88

   87:   await middleware(t.context.req, t.context.res, next);
   88:   helpers.sendErrorResponse.should.have.been.called;   
   89:   next.should.not.have.been.called;                    

  Rejected promise returned by test

  Rejection reason:

    [AssertionError: expected sendErrorResponse to have been called at least once, but it was never called]

```

aaronschachter:
The only routes that will be left after Conversations launch is:
* `POST /receive-message`
* `GET /campaigns`
* `GET /campaigns/:id`

I'd prefer to hold off on making the `req.user._id` change in this branch, and clean it up once we're ready to deprecate all other endpoints (and the User model entirely)

PR #963, opened 2017-09-21
Chore/add first integration test chatbot

rapala61:
#### What's this PR do?
I had this integration test in my stash for a while. I'm just adding it to the repo so we have an starting point to continue to test as much as possible all edge cases in `/chatbot`.

- Run `npm run all-tests`


test/integration/routes/chatbot.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+// app
+const app = require('../../../app');
+
+test.after.always(() => {
+  mongoose.connection.db.dropDatabase();
+});
+
+test('Should fail if request does not include a valid api key', async () => {
+  await request(app)
```
sergii-tkachenko: It's an async function. Maybe it'll be better without using then?

test/utils/stubs.js
```diff
@@ -6,12 +6,18 @@
 const underscore = require('underscore');
 
 module.exports = {
+  getNorthstarAPIBaseUri: function getNorthstarAPIBaseUri() {
+    return process.env.DS_NORTHSTAR_API_BASEURI || 'https://northstar-thor.dosomething.org/v1';
```
sergii-tkachenko: Might be better to default to localhost instead.

test/utils/factories/user.js
```diff
@@ -8,10 +8,10 @@ const stubs = require('../../utils/stubs');
 
 const chance = new Chance();
 
-module.exports.getValidUser = function getValidUser() {
+module.exports.getUser = function getUser(phoneNumber) {
```
sergii-tkachenko: I didn't have time to implement it in blink, but I'd recommend looking into [`factory_girl`](https://github.com/aexmachina/factory-girl). It may be more convenient, as it allows both to generate fields from chance.js and override them on demand (as your're doing here with a phoneNumber).

test/integration/routes/chatbot.test.js
rapala61: You're right! I'm going to use the [.expect](https://www.npmjs.com/package/supertest#expectstatus-fn) function instead.

test/utils/stubs.js
rapala61: I've never user a local Northstar and the `.env.example` uses `https://northstar-fake.dosomething.org/v1` should I use that one?

test/utils/factories/user.js
rapala61: I'll take a look in future integration tests.

rapala61:
@sergii-tkachenko updated!

test/utils/stubs.js
sergii-tkachenko: Yep. The idea is you could use a lib that would fake responses from this domain. Can be really helpful with tests.

test/utils/factories/user.js
sergii-tkachenko: Yep, just a note.

test/utils/stubs.js
rapala61: Ok. I use nock for tests. Updated.

PR #965, opened 2017-09-22
Receive Message: required userId param

### LINKED ISSUE #959: Refactor Receive Message to accept userId parameter

aaronschachter:
#### What's this PR do?
Now that Gambit Conversations is passing Northstar `userId` parameter when available (https://github.com/DoSomething/gambit-conversations/pull/140), and our Signup methods accept a `userId` string instead of a `user` object (#962), we can remove the Create / Get User middleware from `POST /receive-message` to avoid making double Northstar requests in Gambit Conversations.

Can't deploy this yet because we don't want to break Slack / Consolebot integration in the Conversations API - we'll need to figure out what `userId` to pass over for non phone numbers.

Next up:
* [x] in Gambit Conversation, pass `userId` in Campaigns Receive Message requests for Conversations with platforms `'slack'` and `'api'`.
* [ ] Refactor `User.updateConversation` as `User.updateCurrentCampaign`
* [ ] Refactor Receive Message API response to return a Signup and the appropriate reply template, to unblock last task in https://github.com/DoSomething/gambit-conversations/issues/37

#### Relevant tickets
Fixes #959 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/conversation/replies.js
+++ b/lib/conversation/replies.js
@@ -17,7 +17,7 @@ const BotRequest = require('../../app/models/BotRequest');
 
  /**
   * Decorator.
-  * Returns passed function wrapped with functionality to respond with a succes response to
+  * Returns passed function wrapped with functionality to respond with a success response to
   * the user.
   *
   * @param  {function} fn function to be wrapped.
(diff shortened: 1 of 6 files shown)


aaronschachter:
This is now ready for review per https://github.com/DoSomething/gambit-conversations/pull/144

lib/conversation/replies.js
```diff
@@ -33,78 +33,84 @@ function customMsg(fn) {
...
+  return (replyTemplate, args) => helpers.renderMessageTemplate(args.req, replyTemplate)
+      .then((replyText) => {
+        if (!args.req.user) {
+          return helpers.sendResponse(args.res, 200, replyText, replyTemplate);
+        }
 
+        args.replyText = replyText;
         // User's Current Campaign may need to be updated, pass req.campaignId to check.
-        args.req.user.updateConversation(messageTemplate, args.req.campaignId)
+        return args.req.user.updateConversation(replyTemplate, args.req.campaignId)
```
rapala61: I'm wondering why we are updating the Conversation here, but not [above](https://github.com/DoSomething/gambit/pull/965/files#diff-d00b1362d15f78f5fccb7ba59af8dffeL71).

aaronschachter: One of the next tasks defined in this PR description is to refactor `updateConversation` as just `updateCampaign`. We shouldn't need to update the User's `current_campaign` if the User's being directed into Agent View / Support.

The `updateConversation` function was introduced in https://github.com/DoSomething/gambit/pull/934, when we were looking to move all Gambit Conversations logic into this repo, monolith style. We don't have a need to store or update a User's `lastOutboundTemplate` in this repo anymore (especially since we'll be in a place to remove the User model completely once Conversations goes live).

PR #968, opened 2017-09-29
Remove template rendering from Receive Message response

aaronschachter:
#### What's this PR do?
* Refactors `/lib/conversation/replies` functions to avoid rendering the reply message template if the loaded `req.user` doesn't exist. 

   * Unblocks last task in https://github.com/DoSomething/gambit-conversations/issues/37: This means each `POST /receive-message` request will only make a GET request to Phoenix Campaigns, and an additional POST request if we need to create a new Signup or submit a completed draft Reportback Submission.

    * Refactors `POST /receive-message` response object to return `res.body.data.replyTemplate` string instead of `res.body.data.reply` object

* Fixes epic bug in `POST /receive-message` introduced by #965: If a User model doesn't exist for the `userId` we pass in a `POST /receive-message`, the `Signup.postReportbackSubmission` function because it was expecting a populated User model. We're not using those anymore in Receive Message.

* Adds templates for `invalidCaption` and `invalidWhyParticipated` -- new fields I've defined in our Contentful Campaign content type

Will be opening up a branch in Conversations that fixes the things that this would break.

#### Relevant tickets
* Fixes bug from #965 
* Unblocks https://github.com/DoSomething/gambit-conversations/issues/37

#### Checklist
- [x] Tested on staging.


lib/conversation/replies.js
```diff
@@ -40,22 +40,23 @@ function customMsg(fn) {
...
-      stathat.postStat(`campaignbot:${replyTemplate}`);
-      BotRequest.log(args.req, 'campaignbot', replyTemplate, replyText);
-
-      return helpers.sendResponse(args.res, 200, replyText, replyTemplate);
-    })
-    .catch(err => helpers.sendErrorResponse(args.res, err));
+  return (replyTemplate, args) => {
+    if (!args.req.user) {
+      const signup = args.req.signup;
+      return helpers.sendResponseForSignup(args.res, signup, replyTemplate);
```
rapala61: I know this is something it's happening throughout the codebase (wasn't introduced by this commit) and we might have to do cleanup at a later point, but if a function returns a promise at any point, we should promisify all branches of the function, just so we don't end up returning an object to a function that was expecting a promise under some circumstances or not rejecting when that is what we intended to do. I fixed something like that in the code I am working on in [Conversations](https://github.com/DoSomething/gambit-conversations/blob/master/lib/helpers.js#L276) (fix not pushed yet).

aaronschachter: Agreed - this `/lib/conversation` will definitely need some major cleanup once we go live -- this was a workaround to avoid making changes to function calls in all of the various middleware.

Issue #970, opened 2017-10-04
Switch Get/Post Signups & Reportbacks from Phoenix to Rogue


aaronschachter:
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer I'd like to post directly to Rogue instead of the Phoenix proxy endpoints to create Signups and Reportbacks, and query the Rogue API for fetching a user's current Signup.

#### Additional Information (optional)
#922 

#### Tasks
* [x] Post Signup to Rogue via https://github.com/DoSomething/rogue/blob/master/documentation/endpoints/signups.md#signups - #974 
* [x] Post Reportback to Rogue via https://github.com/DoSomething/rogue/blob/master/documentation/endpoints/posts.md#create-a-post-and/or-create/Update-a-signup
* [x] Get current Signup from Rogue via https://github.com/DoSomething/rogue/blob/master/documentation/endpoints/activity.md#activity

PR #974, opened 2017-10-10
Post Signups to Rogue

aaronschachter:
#### What's this PR do?
Starts on Rogue integration, posting Signups to Rogue instead of Phoenix if `DS_ROGUE_ENABLED` config var is set to true.

Set `DS_ROGUE_ENABLED` to `true` and add the relevant config variables for Base URI and API Key. Use consolebot to test signups for a new phone number to verify Signups are created.

Set `DS_ROGUE_ENABLED` to `false` -- verify Signup is posted to Phoenix. this is currently not working on Thor (we get the "API response is false" error -- #devops issue)

Once this is deployed to production and works beautifully, we can remove the conditional posting to Phoenix and always post to Rogue, to keep code simpler.

Next up: posting Reportbacks

Because we're looking to get this in for testing with TSG, haven't added test coverage yet. 🤦‍♂️ 👎 

#### Relevant tickets
First task in #970

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/rogue.js
+++ b/lib/rogue.js
@@ -3,24 +3,40 @@
 /**
  * Imports.
  */
-const logger = require('winston');
+const Promise = require('bluebird');
 const superagent = require('superagent');
 
 const defaultConfig = require('../config/lib/rogue');
 
+/**
+ * @param {string} endpoint
+ * @param {object} data
+ */
 function executePost(endpoint, data) {
-  const uri = `${this.options.baseUri}/${endpoint}`;
+  const url = `${defaultConfig.clientOptions.baseUri}/${endpoint}`;
 
   return superagent
     .post(url)
-    .set('X-DS-Rogue-API-Key', defaultConfig.apiKey)
-    .send(data);
-};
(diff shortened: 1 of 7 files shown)


lib/rogue.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * @param {object} req
+ * @return {Promise}
+ */
+module.exports.postSignupForReq = function (req) {
+  const campaign = req.campaign;
+  const userId = req.userId;
+
+  const data = {
```
rapala61: Would be nice to add the @see for the payload docs here too.

aaronschachter:
I've tested with Rogue production creds locally just to make sure it worked, it did.

Still at a loss for why the post Signup to Phoenix is failing when `DS_ROGUE_ENABLED` is set to `false`. This is happening on `master` too.

aaronschachter:
Phoenix signups fixed per API credentials: https://dosomething.slack.com/archives/C03T8SDJJ/p1507666395000143?thread_ts=1507640453.000332&cid=C03T8SDJJ

PR #991, opened 2017-10-27
Updates node version

aaronschachter:
doge-code


aaronschachter:
Tested a Thor keyword and went through the Campaign flow, looks 👍

PR #992, opened 2017-10-27
WIP: Always return GET /campaigns/:id response if Phoenix Campaign ID exists

### LINKED ISSUE #990: GET /campaigns/:id error: Cannot read property 'fields' of undefined
_BUG_
-------------------
#### Current Behavior
- Seeing a bunch of 500 errors on staging to `GET /campaigns/2003` with a `Cannot read property 'fields' of undefined` error. It's not happening on production.  My hunch is that this is from https://github.com/DoSomething/gambit-conversations/issues/189. [Campaign 2003 is a valid Phoenix Campaign](https://thor.dosomething.org/api/v1/campaigns/2003), but it doesn't have any Keywords defined in Contentful.

#### Desired Behavior
- Either send a 200 back with empty `templates` and `keywords`, or send a 404 error. I'm leaning toward a 200 with empty properties -- curious for your input @rapala61 and @sergii-tkachenko. 

Technically a Campaign could still have templates without a Contentful Campaign defined, we'd just render the default Campaign template values. The lack of keywords would indicate it's not currently available on Gambit -- but begs the question, should it be a 404 if not currently available on Gambit? I think the answer is yes - on the Gambit admin side, we'd want to see the templates and messages for Campaigns no longer available for members to complete.

#### Why This Matters
- 500 seems like the wrong error code to return here, if we should return an error at all.

aaronschachter:
#### What's this PR do?
Resolves #990 by always returning a response regardless of whether a Phoenix Campaign has a corresponding Contentful campaign. If it doesn't have a Contentful Campaign, we return the `templates` property rendered with the default Contentful Campaign data.

I can deploy to GamCam Staging to confirm expected behavior for both the `GET /campaigns/:id` and `POST /chatbot` endpoints, and that we stop seeing the errors in #990 from web signups on Campaigns without keywords -- but this breaks a bunch of tests, which is why it's not ready to merge just yet. It might be easier to review at this point, as it'd just be less lines of code to review... 

Next up in this branch:
* removing or preferably fixing failing tests
* Improving documentation for endpoints:
    * `GET /campaigns` - Returns all Phoenix Campaigns that have published Gambini Keywords (regardless of Phoenix Campaign Status)
    * `GET /campaigns/:id` - Returns to rendered Gambit templates to be used for given Phoenix Campaign ID if it's open on Gambit (has Phoenix Campaign status `active` and at least one published Keyword) 

#### Relevant tickets
Fixes #990 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -258,8 +258,10 @@ module.exports.fetchDefaultAndOverrideContentfulCampaignsForCampaignId = functio
     const campaignIds = [defaultCampaignId, campaignId];
 
     return exports.fetchCampaigns(campaignIds)
-      .then(res => Promise.resolve(parseDefaultAndOverrideContentfulCampaignsResponse(res.items)))
-      .then(data => resolve(data))
+      .then((res) => {
+        const data = parseDefaultAndOverrideContentfulCampaignsResponse(res.items);
+        return resolve(data);
+      })
       .catch(error => reject(exports.contentfulError(error)));
   });
 };
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -103,7 +103,7 @@ module.exports.fetchCampaign = function (campaignId) {
  * @param {array} entries - Assumes the Contentful Campaign with defaultCampaignId is included.
  * @return {object}
  */
-function mergeCampaignEntries(entries) {
+function mergeContentfulCampaignEntries(entries) {
   let defaultContentfulCampaign;
   let overrideContentfulCampaign;
   const templateNames = Object.keys(config.campaignFields);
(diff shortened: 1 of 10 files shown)


lib/contentful.js
```diff
@@ -188,91 +187,64 @@ module.exports.renderMessageForPhoenixCampaign = function (phoenixCampaign, msgT
...
-    // use getEntries() with an IN operator get both Contentful Campaigns in a single request.
-    // @see https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/inclusion
-    return exports.fetchCampaign(campaignId)
-      .then((contentfulCampaign) => {
-        campaignContent.overrides = contentfulCampaign.fields;
-        return exports.fetchCampaign(defaultCampaignId);
-      })
-      .then((defaultContentfulCampaign) => {
-        campaignContent.defaults = defaultContentfulCampaign.fields;
-        templateNames.forEach((templateName) => {
```
aaronschachter: Moved to `/lib/middleware/campaign-single/render-templates.js`

Issue #989, opened 2017-10-25
5.0.0 cleanup


aaronschachter:
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to remove all code that won't be used in production when Conversations is live.

#### Tasks

* [x] Delete `BotRequest` model - #1007 
* [x] Delete `User` model - #1002 
* [x] Delete `logs` winston collection
* [x] Delete `POST /campaign/:id/message`  route - #998
* [x] Delete `POST /chatbot` route - #998 
* [x] Delete `POST /signups` route - #998 
* [x] Move middleware used by `receive-message` into a `receive-message` directory, delete middleware no longer used, e.g. `user-get` - #1009 
* [x] Delete `lib/consolebot` ? - #1001 
* [x] Delete `/lib/groups` - #996 
* [x] Delete `/lib/mobilecommons` - #1007 
* [x] Delete `lib/northstar` - #1004 
* [x] Refactor `lib/conversation` - #1007, #1009 
* [x] Deprecate `lib/middleware/map-request-params` - #1005 
* [x] Remove unused helpers - #1008 
* [x] Move relevant tests into Gambit Campaigns

PR #998, opened 2017-11-01
TGM: Remove deprecated routes

aaronschachter:
#### What's this PR do?
Removes deprecated routes:
* `POST /campaigns/:id/message`
* `POST /chatbot`
* `POST /signups`

Upon staging deploy:
* Confirm `GET /campaigns` and `GET /campaigns/:id` endpoints are 👍 
* Confirm Campaign conversations in Gambit Conversations Staging work as expected

Let's aim to kill the beast PR by PR in hopes of it being easier to test and review? Starting with these routes we no longer post to.

#### Relevant tickets
#989 

#### Checklist
- [x] Tested on staging.


package.json
```diff
@@ -8,7 +8,7 @@
     "url": "git+https://github.com/DoSomething/gambit.git"
   },
   "scripts": {
-    "test": "NODE_ENV=test ava --serial",
+    "test": "NODE_ENV=test ava --serial --fail-fast",
```
aaronschachter: Happy to remove this, but I slightly prefer it?  Could be swayed though.

rapala61: What about adding another script named `test-fast`? I believe there is something to gain from seeing all failed tests across different files.

aaronschachter: Good call! Will update.

PR #1007, opened 2017-11-21
TGM: Replies - part 1

aaronschachter:
#### What's this PR do?

* Removes deprecated Mobile Commons, BotRequest code
* Refactors the Reply Dispatcher accordingly, for sake of not changing all of the Receive Message middleware too. 

I'll deploy this to staging to confirm via Gambit Conversations Staging that `POST /receive-message` requests work as expected.

Thoughts for Part 2: 
* Move `lib/conversation/replies` into `lib/replies`
* Deprecate `lib/conversations/reply-dispatcher` and  `lib/conversations/command`
* Find/replace all middleware calls to return `replies[replyTemplate](req, res)` instead of   `ReplyDispatcher.execute(replies[replyTemplate]( { req, res })`.

#### Relevant tickets
#989 

#### Checklist
- [x] Tested on staging.


lib/conversation/reply-dispatcher.js
```diff
@@ -1,12 +1,9 @@
 'use strict';
 
-const logger = require('winston');
-
-// TODO: Should this be a singleton?
+// TODO: Remove me?
```
rapala61: Sure, if we removed 80% of the abstraction, it makes sense to move the remainder.

PR #1013, opened 2017-12-09
Invalidates decimal reportback quantities.

### LINKED ISSUE #1012: Floating quantities are breaking Posts to Rogue v2
_BUG_
-------------------
#### Current Behavior
- When an user reportback a quantity with decimals, Rogue seems to return a 200 but not a parseable response to Gambit Campaigns.

#### Desired Behavior
- We should be able to process integers or decimal quantities. If there are other limitations we should make sure we can handle them.

##### Observed problem
- Tonight (12/7/17). An user reported back a quantity of `15.9`.
- When submitting the finalized draft to Rogue. Rogue appeared to respond with an empty `200`.
    - No errors in Rogue NewRelic
- Gambit Campaigns attempted to parse this response and produced an error: `error: Signup.createPostForReq error:Cannot read property 'id' of undefined`.
    - This error kept going while investigating.

##### Troubleshooting steps
- Set `LOGGING_LEVEL` to  `debug` in Heroku for Gambit Campaigns.
    - This allowed me to see the payload of the data posted to Rogue:
    ```
    {
	    "data": {
		    "source": "sms",
		    "campaign_id": 7978,
		    "campaign_run_id": 7979,
		    "northstar_id": "5893989ea0bfad67d02d70b9",
		    "quantity": "15.9",
		    "caption": "don’t think of it as quitting. think of it as gaining",
		    "remote_addr": "172.17.65.170",
		    "why_participated": "don’t think of it as quitting. think of it as gaining"
	    }
    }
    ```
- Attempted to manually POST this payload with Postman. Received a `200`, but also html of a login page. Not sure if it was my mistake in this step, but seemed really odd.

##### Steps to fix
- Out of just pure intuition, I went inside Gambit Campaign's database and manually updated this user's reportback submission quantity from `15.9` to `16`.
- This step immediately fixed the problem, as it POSTed successfully on next retry.

#### Why This Matters
- It could happen again at any point. If we are unlucky, we could get swamped with them and it would crash Gambit Campaigns.

rapala61:
##### Part 1
- Refactors reportback specific methods out of the `helpers.js` file into its own helper `/helpers/reportback.js`.
- Makes decimal reportback quantities invalid.

##### Part 2 - in future PR
It will round up/down based on decimal quantity remainder when a valid Numeric quantity is received.

- 👀 
- Run `npm run all-tests`

#### Relevant tickets
Task 1 - https://github.com/DoSomething/gambit-campaigns/issues/1012

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/reportback.js
+++ b/lib/helpers/reportback.js
@@ -1,0 +1,65 @@
+'use strict';
+
+const config = require('../../config/lib/helpers/reportback');
+
+/**
+ * @param {string} message
+ * @return {boolean}
+ */
+function hasLetters(message) {
+  return RegExp(/[a-zA-Z]/g).test(message);
+}
+
+/**
+ * TODO: Make this much better!
+ * @see https://github.com/DoSomething/gambit/issues/608
+ *
+ * Checks if given input string contains a valid Reportback quantity.
+ * @param {string} input
+ * @return {boolean}
+ */
+function validateQuantity(quantity) {
+  // Letters are invalid
+  if (hasLetters(quantity)) return false;
+
+  const numericQuantity = Number(quantity);
+
+  if (numericQuantity && numericQuantity === parseInt(numericQuantity, 10)) {
+    return true;
+  }
+
+  return false;
+}
+
+/**
+ * Checks if given input string contains a valid Reportback text field.
+ * @param {string} input
+ * @return {boolean}
+ */
+function validateText(input) {
+  return !!(input && input.trim().length > config.minTextLength && hasLetters(input));
+}
+
+/**
+ * Trims given input with an ellipsis if its length is greater than 255|maxTextLength characters.
+ * @param {string} input
+ * @return {string}
+ */
+function trimText(input) {
+  const maxLength = config.maxTextLength;
+  const ellipsis = config.ellipsis;
+  const keepLength = maxLength - ellipsis.length;
+  let text = input.trim();
+
+  if (text.length > maxLength) {
+    text = `${text.substring(0, keepLength)}${ellipsis}`;
+  }
+
+  return text;
+}
+
+module.exports = {
+  isValidQuantity: qty => validateQuantity(qty),
+  isValidText: txt => validateText(txt),
+  trimText: txt => trimText(txt),
+};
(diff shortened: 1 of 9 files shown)


lib/middleware/receive-message/draft-quantity.js
```diff
@@ -13,7 +14,8 @@ module.exports = function draftSubmissionQuantity() {
       return replies.askQuantity(req, res);
     }
 
-    if (!helpers.isValidReportbackQuantity(req.incoming_message)) {
+    // TODO: Round up/down based on remainder on Numeric quantity
```
rapala61: TODO for part 2 of #1012.

lib/helpers/reportback.js
```diff
@@ -0,0 +1,65 @@
...
+ * @return {string}
+ */
+function trimText(input) {
+  const maxLength = config.maxTextLength;
+  const ellipsis = config.ellipsis;
+  const keepLength = maxLength - ellipsis.length;
+  let text = input.trim();
+
+  if (text.length > maxLength) {
+    text = `${text.substring(0, keepLength)}${ellipsis}`;
```
aaronschachter: @DFurnes @sbsmith86 -- We're currently trimming Caption and Why Participated if they have lenght greater than 255 characters - due to posting to the Ashes endpoint and #881. Is this still required?

sbsmith86: @aaronschachter `caption` is still limited to 255 but we can update that. `why_participated` is a `TEXT`field so it shouldn't be a concern.

DFurnes: I think it might still make sense to restrict length on captions (since we have to then show them in the gallery w/ a limited amount of space). We validate [4-80 characters](https://github.com/DoSomething/phoenix-next/blob/ca5ef704046b93e233449d7cd11a5c9acf3d8544/app/Http/Controllers/ReportbackController.php#L44) in Phoenix Next, for example.

rapala61: Hmm.. so should we trim text past 60(?) characters for captions? Right now its trimming to 255 characters. @DFurnes ? I'm going to remove the trimming on the why participated field based on your feedback @sbsmith86 @DFurnes.

DFurnes: I don't think it's an absolute requirement, but it does make it easier to display captions in the UI without having to truncate text. IMO, it'd make sense to enforce the same 80 character limit everywhere.

Issue #1003, opened 2017-11-13
Cache active campaigns in Redis.


rapala61:
_OVERVIEW_
-------------------
#### User Story
When the Conversations API sends a GET request to the `/campaigns` route, the response should be cached in Redis, so that any subsequent request gets a cached response until it expires or is removed manually from the Redis cache.

#### Why This Matters
Caching makes our infrastructure more efficient and resilient to momentary platform failures (in theory).

#### Related Issues

Convo [issue#217](https://github.com/DoSomething/gambit-conversations/issues/217)

rapala61:
Elevating this issue to high priority. We just had a pretty severe outage in our Rogue platform that also took down Phoenix. If we had these campaigns cached in Redis, we could have been operational to non reportback/signup texts. Unfortunately, all incoming user texts are failing at this moment. [Slack thread](https://dosomething.slack.com/archives/C0YGXUE01/p1515606077000535)

cc @aaronschachter @justkika

aaronschachter:
Refs https://github.com/orgs/DoSomething/teams/team-odoyle/discussions/2

aaronschachter:
Closed in #1019

PR #1019, opened 2018-03-16
Support Ashes and Phoenix URLs via config 

aaronschachter:
#### What's this PR do?

This PR takes a stab at changing the Phoenix API from old (PA) to new (PN), and seems to find a blocker. The index `GET /campaigns` resource Gambit uses to get a list of Campaigns by given `ids` query parameter doesn't exist. As a workaround, this PR makes multiple requests for each Campaign -- but it seems awfully slow on localhost. We may need to significantly refactor how we manage the list of active Campaigns, or per [Slack comment](https://dosomething.slack.com/archives/C3ASB4204/p1521047195000687?thread_ts=1521038744.000837&cid=C3ASB4204) from @DFurnes, get that [GraphQL](https://github.com/dosomething/graphql) in:
>  I'd go with multiple requests (or see about getting this into GraphQL) so we can more easily cache individual resources.

Update your `DS_PHOENIX_API_BASEURI` to the new Phoenix URL and verify data is confirmed. Note -- some properties are now missing, see background context I want to provide 👇 

Yes! Two things:

#### 1. Breaking changes
The data we use for the `{{problem_fact}}`, `{{rb_confirmation_msg}}`, `{{rb_noun}}`, and `{{rb_verb}}` tags no longer exists. Once we're ready to merge and deploy this, we'll need to make a few of the Gambini Campaign override fields required, like the Signup Menus, Ask Quantity, Photo, etc.

#### 2. List of active campaigns 
The Gambit Campaigns `GET /index` returns all campaigns that are currently available on Gambit (because they have keywords published). This is used for:

* the MENU command, which finds a random campaign to ask User to signup for

* Our admin apps (Gambit Admin, the Gambit Slack app) return the list for staff viewing

I'm very open to rethinking how this could work if there are any ideas... 

cc @weerd 

#### Relevant tickets
First 👶👣 for https://www.pivotaltracker.com/story/show/152693729 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers/campaign.js
+++ b/lib/helpers/campaign.js
@@ -18,14 +18,18 @@ module.exports = {
    * @return {Boolean}
    */
   isClosed: function isClosed(campaign) {
-    const closedValue = config.statuses.closed;
     if (campaign.status) {
-      return campaign.status === closedValue;
+      return this.isClosedStatus(campaign);
     }
+
     if (!campaign.endDate) {
       return true;
     }
+
     return this.isEndDatePast(campaign);
+  },
+  isClosedStatus: function isClosedStatus(campaign) {
+    return campaign.status === config.statuses.closed;
   },
   /**
    * @param {Object} campaign
--- a/lib/helpers/campaign.js
+++ b/lib/helpers/campaign.js
@@ -23,7 +23,7 @@ module.exports = {
     }
 
     if (!campaign.endDate) {
-      return true;
+      return false;
     }
 
     return this.isEndDatePast(campaign);
@@ -47,6 +47,8 @@ module.exports = {
    */
   parseCampaign: function parseCampaign(campaign) {
     const campaignId = Number(campaign.legacyCampaignId);
+    logger.debug('parseCampaign', { campaignId });
+
     const result = {
       id: campaignId,
       title: campaign.title,
@@ -55,7 +57,8 @@ module.exports = {
       currentCampaignRun: { id: Number(campaign.legacyCampaignRunId) },
       endDate: campaign.endDate,
     };
-    logger.debug('parseCampaign', { campaignId });
+    logger.debug('parseCampaign', { result });
+
     return result;
   },
 };
--- a/lib/helpers/campaign.js
+++ b/lib/helpers/campaign.js
@@ -42,10 +42,14 @@ module.exports = {
     return config.statuses.active;
   },
   /**
-   * @param {object} campaign
+   * @param {Object} campaign
+   * @param {Boolean} useAshes
    * @return {Object}
    */
-  parseCampaign: function parseCampaign(campaign) {
+  parseCampaign: function parseCampaign(campaign, useAshes) {
+    if (useAshes) {
+      return this.parseAshesCampaign(campaign);
+    }
     const campaignId = Number(campaign.legacyCampaignId);
     logger.debug('parseCampaign', { campaignId });
 
@@ -58,6 +62,32 @@ module.exports = {
       endDate: campaign.endDate,
     };
     logger.debug('parseCampaign', { result });
+
+    return result;
+  },
+  /**
+   * @param {object} data
+   * @return {Object}
+   */
+  parseAshesCampaign: function parseAshesCampaign(data) {
+    const result = {};
+    const languageCode = data.language.language_code;
+    result.id = Number(data.id);
+    result.title = data.title;
+    result.tagline = data.tagline;
+    result.status = data.status;
+    result.currentCampaignRun = {
+      id: Number(data.campaign_runs.current[languageCode].id),
+    };
+    const rbInfo = data.reportback_info;
+    result.reportbackInfo = {
+      confirmationMessage: rbInfo.confirmation_message,
+      noun: rbInfo.noun,
+      verb: rbInfo.verb,
+    };
+    result.facts = {
+      problem: data.facts.problem,
+    };
 
     return result;
   },
--- a/lib/helpers/campaign.js
+++ b/lib/helpers/campaign.js
@@ -46,15 +46,16 @@ module.exports = {
    * @return {Object}
    */
   parseCampaign: function parseCampaign(campaign) {
+    const campaignId = Number(campaign.legacyCampaignId);
     const result = {
-      id: Number(campaign.legacyCampaignId),
+      id: campaignId,
       title: campaign.title,
       tagline: campaign.tagline,
       status: this.parseStatus(campaign),
-      currentCampaignRun: Number(campaign.legacyCampaignRunId),
+      currentCampaignRun: { id: Number(campaign.legacyCampaignRunId) },
       endDate: campaign.endDate,
     };
-    logger.debug('parseCampaign', result);
+    logger.debug('parseCampaign', { campaignId });
     return result;
   },
 };
--- a/lib/helpers/campaign.js
+++ b/lib/helpers/campaign.js
@@ -3,13 +3,14 @@
 const dateFns = require('date-fns');
 const logger = require('winston');
 const config = require('../../config/lib/helpers/campaign');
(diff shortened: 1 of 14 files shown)


lib/helpers/campaign.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+const dateFns = require('date-fns');
```
rapala61: Just wondering why the choice of `date-fns` over the seemingly more popular `moment` [module](https://www.npmjs.com/package/moment).

aaronschachter: [Code was stolen from Phoenix Next](https://github.com/DoSomething/phoenix-next/blob/e6711c9330c026e30321eaa4980a093b83df9cf7/resources/assets/helpers/index.js#L331) 🕵️

lib/helpers/campaign.js
```diff
@@ -0,0 +1,60 @@
...
+    if (this.isClosed(campaign)) {
+      return config.statuses.closed;
+    }
+    return config.statuses.active;
+  },
+  /**
+   * @param {object} campaign
+   * @return {Object}
+   */
+  parseCampaign: function parseCampaign(campaign) {
```
rapala61: Minor suggestion. You could start leaving generous amounts of `//TODO`s for when we switch to use varchar campaign ids.

lib/phoenix.js
```diff
@@ -40,33 +41,6 @@ function executeGet(endpoint, query = {}) {
     .then(res => res.body);
 }
 
-/**
- * @param {object} data
- * @return {object}
- */
-module.exports.parsePhoenixCampaign = function (data) {
```
rapala61: Nice cleanup 👍

lib/phoenix.js
```diff
@@ -130,13 +103,14 @@ module.exports.getCampaignById = function getCampaignById(campaignId) {
...
-  logger.debug(`phoenix.fetchCampaigns:${campaignIds}`);
-  const endpoint = `campaigns?ids=${campaignIds.join(',')}`;
-
-  return new Promise((resolve, reject) => {
-    executeGet(endpoint)
-      .then(res => resolve(res.data.map(campaign => exports.parsePhoenixCampaign(campaign))))
-      .catch(err => reject(exports.parsePhoenixError(err)));
+  const promises = [];
+  campaignIds.forEach((campaignId) => {
+    promises.push(module.exports.getCampaignById(campaignId));
```
rapala61: This looks a little cringe inducing at first, but since we are [caching each one of these requests](https://github.com/DoSomething/gambit-campaigns/blob/master/lib/phoenix.js#L121) in Redis, once the cache is warmed up in one request, all other calls will quickly return the cached campaigns.

aaronschachter: Hmm, well I can try deploying to staging... if it would be ok then this wouldn't be blocked!

rapala61: Yea, let's test performance on staging before unblocking. I'm optimistic the cache should make it bearable.

lib/helpers/campaign.js
weerd: @rapala61 the reason specifically is that `date-fns` has a significantly smaller footprint on the code base.

aaronschachter:
I'm going to give a staging deploy a shot. Will change the `DS_PHOENIX_API_BASEURI` to the example Phoenix URL and add overrides for some of our usual test campaigns that don't have any (2299, 2900 are my usual suspects).

aaronschachter:
This has been deployed to staging and passes functional tests. Blocked by:

* [ ] Load testing. @rapala61 - any suggestions for best approach here? Simulating a broadcast from Conversations, or potentially just using Siege?

* [ ] Phoenix data is in correct for some campaigns, 8017 is returning the wrong title. See [Slack thread,](https://dosomething.slack.com/archives/C782XBKDM/p1521477975000843?thread_ts=1521218968.000004&cid=C782XBKDM)

* [ ] Gambini updates. Our Campaign content type will need to make certain fields that use missing tags required, e.g. `askQuantity`, invalidQuantity, `askPhoto`, `invalidPhoto` and we'll need to make sure all Campaigns currently live have values set in these fields.

lib/helpers/campaign.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     }
 
     if (!campaign.endDate) {
-      return true;
+      return false;
```
aaronschachter: Now that the response for [Grab the Mic](https://next.dosomething.org/api/v2/campaigns/8017) is looking better, I realized if the `endDate` is null, status should be parsed as `'active'`.

(although.. the `headline` in all caps may come across as quite shouty if we don't override the campaign `askSignup` template)

lib/phoenix.js
aaronschachter: Testing locally -- when I flush redis cache and `GET /index` with `NODE_ENV` set to `production`, I do get an Internal Server Error that I'm trying to track down. The subsequent `GET /campaigns` request is successful, as some campaigns are returned from cache.

aaronschachter:
For now, I'm going to edit the [default Gambini campaign template](https://app.contentful.com/spaces/owik07lyerdj/entries/3gWKLwCIPmouSosiUkGkOi) to remove any references to tags we no longer support (e.g. noun, verb, problem fact, etc), instead of changing these to fields mandatory (for now), to resolve the Gambini content blocker.

I've rebased this branch with the latest `master` and will deploy to staging to test the reportback flow after I've made the edits to the [default template](https://app.contentful.com/spaces/owik07lyerdj/entries/3gWKLwCIPmouSosiUkGkOi) in Contentful.

aaronschachter:
This branch has been updated to support both Ashes and Phoenix per [Slack thread](https://dosomething.slack.com/archives/C03T8SDJJ/p1522258083000122?thread_ts=1522085399.000668&cid=C03T8SDJJ) on merging and deploying this PR.

When the  `GAMBIT_CAMPAIGNS_SOURCE` var is set to `'ashes'` -- we query the Ashes API. I've deployed to staging, set config vars, tested, and confirmed the responses work as expected. Before a production deploy, we will need to add the following config vars:
* [ ] `GAMBIT_CAMPAIGNS_SOURCE` set to `'ashes'`
* [ ] `DS_CAMPAIGNS_API_BASEURI` set to Phoenix URL
* [ ] `DS_ASHES_API_BASEURI` set to Ashes URL

After production deploy:
* [ ] Unset now deprecated `DS_PHOENIX_API_BASEURI`

aaronschachter:
I think this might close #1003? We're no longer using the Ashes API's `GET /campaigns` resource, and instead making individual requests to `GET /campaigns/:id` -- each of which will be cached after the first request.

test/lib/lib-helpers/campaign.test.js
```diff
@@ -0,0 +1,119 @@
...
+});
+
+test('parseStatus returns closed status value if campaign isClosed', () => {
+  sandbox.stub(campaignHelper, 'isClosed')
+    .returns(true);
+  campaignHelper.parseStatus(campaign).should.equal(config.statuses.closed);
+});
+
+// parseCampaign
+test('parseCampaign validations', (t) => {
```
rapala61: Suggestion about test description naming. Something a little more descriptive could have helped me when reading this test. Example: `validate phoenix campaign parsed properties`

aaronschachter:
I ran through another staging deploy per latest changes, and verified both Ashes and Phoenix URL's work. To make the switch from Ashes to Phoenix, we just unset the `GAMBIT_CAMPAIGNS_SOURCE` config var (or set it to `'ashes'` to change back)  A `GET /campaigns/:id` response will contain an `endDate` property if we're querying Phoenix instead of Ashes.

I've also added [the three new config vars to productio](https://github.com/DoSomething/gambit-campaigns/pull/1019#issuecomment-377106822)n, so this should be good to go 🚗

Issue #1021, opened 2018-03-26
New botConfig option to accept text reportbacks


aaronschachter:
_FEATURE OVERVIEW_
-------------------
#### User Story
As an admin, I want the option to configure a campaign to collect and submit a Rogue post with type `'text'`, instead of current type `'photo'`.  

#### Links
* https://www.pivotaltracker.com/story/show/155988891
*  [Spec](https://docs.google.com/document/d/1VjpgRTeVakbbSr1WwkR4tEjXiiVZ9i1bAEkNInsWQrk/edit?usp=sharing)

#### Tasks

*Contentful:*
* [x] Create a new `textPostTemplate` content type in Gambini, with multi-line text fields `askTextMessage` and `invalidTextMessage`
* [x] Create a new `postTemplate` field on a Gambini campaign, as a reference field accepting only `textPostTemplate` entries.

*GET /campaigns:*
* [x] Add `askText` and `invalidText` properties to response `templates` object

*POST /receive-message:*

If Campaign has `postTemplate` set to a `textPostTemplate`:
*  [x] Send response with `replyTemplate` set as  `'invalidText'` if the message text is not valid or contains only emoji
*  [x] if valid, create a signup and post with type `'text'`  and send response with `replyTemplate` set as `'completedMenu'`

#### Additional information

We'll want to create a `photoPostTemplate` content type and migrate the relevant Campaign fields like `askPhoto`, `askQuantity`. Not needed for launching support for text campaign posts.

aaronschachter:
Final task resolved in https://github.com/DoSomething/gambit-campaigns/pull/1035.

PR #1028, opened 2018-04-04
campaignActivity helper [pr]

### LINKED ISSUE #1018: Accept Conversations.platform as Rogue source
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to use the value passed from a `platform` or `source` parameter in a `POST /receive-message` request, instead of hardcoding `'sms'`.

#### Why This Matters
While testing Gambit Staging on Slack, I expected my Reportback source to be set to `'slack'`. It wasn't, it's set to `'slack'` because of said hardcoding.

We'd need this in place for if/when we add Gambit FB Messenger integration.

aaronschachter:
#### What's this PR do?
Removes parsing `req` variables out of the Rogue helper and into a new `campaignActivity` helper, consuming the body parameters passed from Conversations in https://github.com/DoSomething/gambit-conversations/pull/309

* Removes `lib/middleware/receive-message/phoenix-campaign-get`, as we're being sent all the variables in the post body. This decouples `POST /receive-message` from the `GET /campaigns` campaign cache, making it easier to change the source of botConfigs to Phoenix one day instead of Gambini.

* Removes passing the ancient `ip_address` property when creating a post. This is a relic from the Mobile Commons + dawn of Ashes days, before a `source` property existed and we'd identify SMS reportbacks via the Mobile Commons IP address. cc @DFurnes @mshmsh5000 
 
* Passes the `req.platform` as source, so Slack signups and reportbacks will no longer be set to `'sms'`.

Verify signup and reportback conversations work as expected.

There's room to go with more tests and decoupling (like in the Signup model), but opening this for now to help keep reviewable. There are a bazillion line changes because of the `package-lock.json`, we're removing the `ip` package.

#### Relevant tickets
Fixes #1018 
#1021 

#### Checklist
- [x] Tested on staging.


.env.example
```diff
@@ -1,30 +1,37 @@
-NODE_ENV=thor
-
 GAMBIT_API_KEY=totallysecret
 GAMBIT_CMD_MEMBER_SUPPORT=Q
 GAMBIT_CMD_REPORTBACK=P
 
+##
```
rapala61: Pretty commenting in this file 👍

PR #1053, opened 2018-06-06
Return templates when topic doesn't have a campaign

aaronschachter:
#### What's this PR do?
Modifies parsing the response of a `GET /topics/id` request to return a `templates` property, even if the topic doesn't have a `campaign` reference set.

* Adds check for when the `campaignId` field on a `campaign` entry doesn't exist in Phoenix, which is currently happening on `GET /v1/topics/5fpmfk4ETe4ss6qogEcGSi`

* 👀
* Set env vars to Contentful Preview API and verify `GET v1/topics` and `GET /v1/topics/5fpmfk4ETe4ss6qogEcGSi` return success responses

Preview API is fun because Contentful creates an ID as soon as an editor opens the form to create a new entry. I ran into this error when working on https://github.com/DoSomething/gambit-conversations/pull/328 per continuing https://github.com/DoSomething/gambit-conversations/pull/336

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -27,24 +27,28 @@ module.exports.addBlinkSuppressHeaders = function addBlinkSuppressHeaders(res) {
 /**
  * Replaces mustache tags used in raw Campaign Template strings with given variables.
  * @param {string} input
- * @param {object} phoenixCampaign
+ * @param {object} campaign
  * @return string.
  */
-module.exports.replacePhoenixCampaignVars = function (input, phoenixCampaign) {
+module.exports.replacePhoenixCampaignVars = function (input, campaign) {
   if (!input) {
     return '';
   }
 
   let scope = input;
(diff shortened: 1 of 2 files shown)


rapala61:
You probably already noticed, but wanted to flag. Since the campaign doesn't exist. The `rendered` property contains the same raw mustache templates. 
```
        "startPhotoPost": {
          "raw": "Thanks for signing up for {{title}}! Text {{cmd_reportback}} to submit a post.",
          "override": false,
          "rendered": "Thanks for signing up for {{title}}! Text START to submit a post."
        },
```

aaronschachter:
Yup - that's intentional for now. Users shouldn't receive any messaging from a topic without a campaign -- as we save a campaign to their conversation, and lookup the topic to use via `GET /campaigns/:id` -- which has a topics array property of nested topics for the given campaign id.

PR #1055, opened 2018-06-21
Use defaultTopicTriggers instead of keywords for GET /campaigns data [pr]

aaronschachter:
#### What's this PR do?

Sources the list of active campaigns (and active topics for a campaign) from all `defaultTopicTriggers` that change topics. I've created and published new `defaultTopicTriggers` entries for all of our currently published `keyword` entries, so the results of `GET /campaigns` in each branch should be the same (except for sorting)

* Also adds a `name` property per new field I've manually added and populated for current `photoPostConfig`, `textPostConfig`, and `externalPostConfig` entries

* Upon staging deploy, the `GET /campaigns` response should have the same campaigns returned as a production GET request. 
* For testing purposes, one could unpublish a current published `defaultTopicTrigger` with a topic entry (e.g. mirror), flush redis, and then execute the GET /campaigns request -- the campaign of the topic of the defaultTopicTrigger should be removed

This for sure could use some unit tests, which I may add in this branch -- but wanted to open as is to potentially deploy today and start on some major cleanup (removing all references to the keyword entries across repos). I'll open a Gambit Admin PR in tandem that I've been using to spot-check this branch.

#### Checklist
- [x] Tested on staging.


aaronschachter:
I've deployed this updated branch to staging, and set the environment variables to use the Contentful Preview API. I've also deployed https://github.com/DoSomething/gambit-admin/pull/54 to Gambit Admin, as it helps to verify all looks ok -- we're supporting both keywords and defaultTopicTrigger entries at the time being, giving us room to rollback in case of sadness. Once all looks good, we'll remove references to `keywords` completely in both apps to avoid confusion for conversation entry points.

The one "breaking" but not so breaking change this makes in Conversations is that users can no longer unpause support conversations by texting in a keyword. @rapala61 and I chatted about potentially adding properties to `defaultTopicTrigger` entries to determine on whether it should unpause a support conversation, but decided it sounded simpler and logical to stop pausing support conversations altogether - per https://www.pivotaltracker.com/story/show/157987170

You can verify that the defaultTopicTrigger is changing user's topic by checking that its inbound message macro is set to `changeTopic`. @classicfreddie has added a keyword entry since I last tested this so there's not a defaultTopicTrigger entry for it yet. This helps test that keywords are still working -- the conversation topic is changed via keyword, the message macro is set to `catchAll`.


screen shot 2018-06-25 at 9 29 29 am

lib/helpers/defaultTopicTrigger.js
```diff
@@ -12,8 +12,11 @@ const config = require('../../config/lib/helpers/defaultTopicTrigger');
 function fetchAll() {
   // TODO: Loop through all pages of query results - currently returns 100.
   return contentful.fetchByContentTypes(config.defaultTopicTriggerContentTypes)
-    .then(contentfulEntries => contentfulEntries.map(module.exports
-      .parseDefaultTopicTriggerFromContentfulEntry));
+    .then(contentfulEntries => Promise.all(contentfulEntries.map(module.exports
```
rapala61: Minor, but we could make use of the [`Promise.map`](http://bluebirdjs.com/docs/api/promise.map.html) function here to simplify this a little more.

lib/helpers/defaultTopicTrigger.js
```diff
@@ -12,8 +12,11 @@ const config = require('../../config/lib/helpers/defaultTopicTrigger');
...
+      .parseDefaultTopicTriggerFromContentfulEntry)))
+    // Remove any draft entries that may have been parsed as null.
+    .then(defaultTopicTriggers => defaultTopicTriggers
+      .filter(defaultTopicTrigger => defaultTopicTrigger));
```
rapala61: It would make it a little more readable if we could move this filtering to a function instead.

rapala61:
I saw this in the `/campaigns` response and I wonder if having 2 topics with the same id in the `topics` array is intended or a 🐛 .

screen shot 2018-06-25 at 1 05 30 pm

> The *red* is the staging response, the *green* is the production.

aaronschachter:
Great eyes re: https://github.com/DoSomething/gambit-campaigns/pull/1055#issuecomment-400053021 -- that is a bug! It isn't critical to fix, as Conversations [inspects the first topic in a campaign's `topics` array](https://github.com/DoSomething/gambit-conversations/blob/2.11.5/lib/middleware/messages/member/topic-get.js#L30) -- but will be once we support multiple topics. I'll see if it's a quick fix when implementing the other (very good) suggestions.. but if not may fix in a next patch release to finally get this to prod and remove keywords for reals.

PR #1061, opened 2018-07-20
Adds meta property into index query result [pr]

aaronschachter:
#### What's this PR do?
Adds a `meta` object property to the result in addition to the `data` array property for all index endpoints

* Adds support for a `skip` query parameter to `GET /broadcasts` and `GET /topics`.

* No longer caches the `GET /topics` index query per https://github.com/DoSomething/gambit-conversations/pull/371. Eventually we'll accumulate enough topics that we'll want to page through results... caching pages of results isn't required and adds additional complexity for something we don't need.

Verify that passing `skip` query parameters to the `GET /broadcasts` and `GET /topics` endpoints.
 
We still need a cached list of all topics to easily find topics by campaign.. but ideally this should get refactored by creating separate `contentful.fetch` functions to make Contentful queries that find topics by the nested `campaignId` field. I didn't go this route at first because we can't execute a single Contentful query to find by fields over multiple content types, which we need to find topics  by campaign id. An appealing alternative to making the multiple Contentful requests per content type is looking into moving these endpoints into [GraphQL](https://github.com/dosomething/graphql) to make it easier to link entities.  

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158080999

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -120,10 +120,10 @@ function fetchByContentTypes(contentTypes, queryParams = {}) {
     .build();
 
   return module.exports.getClient().getEntries(query)
-    .then((res) => {
-      res.meta = utilHelper.getPagination(res.total, res.skip, res.limit);
-      return res;
-    })
+    .then(res => ({
+      meta: utilHelper.getMeta(res.total, res.skip, res.limit),
+      data: res.items,
+    }))
     .catch((error) => {
       throw module.exports.contentfulError(error);
     });
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -103,22 +103,19 @@ function fetchByContentfulId(contentfulId) {
 
 /**
  * Returns Contentful entries with contentType in given contentTypes array.
- * TODO: Add pagination support, this only returns the first 100 (Contentful default limit).
  *
+ * @param {Number} skip
  * @param {Array} contentTypes
  * @return {Promise}
  */
-function fetchByContentTypes(contentTypes) {
+function fetchByContentTypes(contentTypes, skip = 0) {
   logger.debug('fetchByContentTypes', { contentTypes });
   const query = module.exports.getQueryBuilder()
     .contentTypes(contentTypes)
+    .skip(skip)
     .orderByDescCreatedAt()
     .build();
   return module.exports.getClient().getEntries(query)
-    .then((res) => {
-      logger.debug('fetchByContentTypes count', res.total);
-      return res.items;
-    })
     .catch((error) => {
       throw module.exports.contentfulError(error);
     });
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -10,6 +10,7 @@ const underscore = require('underscore');
 const NotFoundError = require('../app/exceptions/NotFoundError');
 const config = require('../config/lib/contentful');
 const Builder = require('./contentfulQueryBuilder');
+const utilHelper = require('./helpers/util');
 
 const ERROR_PREFIX = 'Contentful:';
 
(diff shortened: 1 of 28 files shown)


aaronschachter:
Tested on staging, `skip` parameter works as expected: http://ds-mdata-responder-staging.herokuapp.com/v1/broadcasts?skip=200

lib/contentful.js
```diff
@@ -103,22 +104,26 @@ function fetchByContentfulId(contentfulId) {
...
     .orderByDescCreatedAt()
     .build();
+
   return module.exports.getClient().getEntries(query)
-    .then((res) => {
-      logger.debug('fetchByContentTypes count', res.total);
-      return res.items;
-    })
+    .then(res => ({
+      meta: utilHelper.getMeta(res.total, res.skip, res.limit),
```
rapala61: I think there is a good opportunity to make this response more declarative by calling a parsing function here:
```
// implementation
function parseContentfulResponse(res) {
    return {
        data: res.items,
        meta: utilHelper.getMeta(res.total, res.skip, res.limit),
    };
}

// usage
.then(res => module.exports.parseContentfulResponse(res))
```

rapala61:
Great job on this first pass at adding pagination metadata. Specially given how complex these implementation are currently. My observation about this pass would be mostly focused around the responding with `meta` and `data` objects, and how they are implemented.

It took me quite a bit of time to wrap my head around why we are building the response's `meta` and `data` objects differently, per helper (topics, broadcasts, defaultTopicTriggers).

Example: The `/broadcasts/index/broadcasts-get.js` middleware gets both the `meta` and `data` directly from the `helpers.broadcast.fetch` method. Whereas in `/defaultTopicTriggers/index/defaultTopicTriggers-get.js` gets its `data` from the `helpers.defaultTopicTrigger.getAll` helpers instead of `fetch`, but builds its `meta` object in the middleware from the length of `data`.

I get that these helpers are very complex so we should make a card to maybe try to simplify and standardize how we implement the contentful collections' fetch, parse, and response. 🔨 🏗 🙂

aaronschachter:
> why we are building the response's `meta` and `data` objects differently, per helper (topics, broadcasts, defaultTopicTriggers).

The broadcasts and topics middleware both send through the meta data from Contentful results but the outlier is the `defaultTopicTriggers` index middleware.  We currently need this route to send through the cached array of all parsed `defaultTopicTrigger` entries because Conversations pings it upon an app restart to write Rivescript. When caching the `defaultTopicTrigger` entry results, we're currently filtering these results to return entries with required fields (this filtering is a result of using Preview mode. Contentful creates a new entry ID anytime you open the webform.)  

We could store the meta data along with the cached parsed entries, but it'd mean to manually clear cache upon deploy to avoid needing to check whether the cache has our entries saved in a `data` property... but I'm leaning towards no longer caching this index endpoint and instead returning parsed triggers as a new `GET /rivescripts` endpoint per https://github.com/DoSomething/gambit-campaigns/pull/1061#discussion_r204489192) -- and definitely for a future PR!

lib/middleware/campaigns/index/response.js
```diff
@@ -36,6 +37,7 @@ module.exports = function sendResponse() {
       return formattedCampaign;
     });
 
-    return res.send({ data });
+    const meta = helpers.util.getMeta(data.length);
```
aaronschachter: Ah -- this is another outlier per https://github.com/DoSomething/gambit-campaigns/pull/1061#issuecomment-407223840. We don't fetch this list from Contentful, but instead from the cached list of topics, filtering by campaign id -- which is why we're inserting `meta` information here. Supporting this `GET /campaigns` endpoint has been more of convenience for Gambit Admin, but should be ideally be deprecated by passing a `campaignId` parameter to our `topics` endpoint -- or better yet, exposing our Contentful index endpoints to GraphQL to DRY

aaronschachter:
> The broadcasts and topics middleware both send through the meta data from Contentful results but the outlier is the defaultTopicTriggers index middleware.

Oops - I forgot that `campaigns` index also adds the `meta` for sake of returning cached data. I'll update the middleware with comments to document why it's going on, but feeling like GraphQL would be the way to go per

> we should make a card to maybe try to simplify and standardize how we implement the contentful collections' fetch, parse, and response

cc @DFurnes

aaronschachter:
Great 👀 on flagging the vagueness around why we sometimes create the `meta` property within the middleware @rapala61 ... I've refactored the `helpers.response.sendDataWithMeta` function as `sendIndexData` aiming to make the code more readable, as well as DRY creating the `meta` object within `sendIndexData` if a `meta` arg is not passed (which happens for the `GET /campaigns` and `GET /defaultTopicTriggers` index middleware -- they return cached data instead of a parsed Contentful `getEntries` response, where the `meta` property would be present).

lib/helpers/defaultTopicTrigger.js
```diff
@@ -8,12 +8,12 @@ const helpers = require('../helpers');
...
+ * @param {Object} query
  * @return {Promise}
  */
-function fetchAll() {
-  // TODO: Loop through all pages of query results - currently returns 100.
-  return contentful.fetchByContentTypes(config.defaultTopicTriggerContentTypes)
-    .then(contentfulEntries => Promise.map(contentfulEntries, module.exports
+function fetch(query = {}) {
+  return contentful.fetchByContentTypes(config.defaultTopicTriggerContentTypes, query)
+    .then(contentfulRes => Promise.map(contentfulRes.data, module.exports
```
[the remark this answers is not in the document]
aaronschachter: Yep, it's declared in line 4 of this file.

PR #1063, opened 2018-07-25
Adds autoReplyBroadcast and askChangeTopicBroadcast types [pr]

aaronschachter:
#### What's this PR do?

A first stab at splitting out our `broadcast` Contentful content type into separate content types taking on 2 of them. We'll pass these Contentful ID to a `POST /messages?origin=broadcast` request as the `broadcastId` -- but Conversations will also save the ID as the conversation topic.  These content types contain a `broadcast` field that references  `message` entries, which we use to send a broadcast message -- but also contain additional fields to return as `templates` for when the user's topic is set to one of these broadcast types.

### autoReplyBroadcast
Deprecates use of the `survey_response` topic broadcasts we currently send.

Fields:

* `name` - text field 
* `broadcast` -the outbound message to send, limited to `message` entries
* `autoReply` - auto reply message to send, also limited to `message` entries

### askChangeTopicBroadcast
Deprecates use of the `askSignup` campaign broadcasts we currently send.

Fields:

* `name`
* `topic` - This is the topic we'll change the conversation to if the user says yes to the broadcast
* `broadcast` - This is the outbound message that will be sent to user, asking them if they'd like to change current topic to this one. If topic is associated with a campaign, this will create a signup 
* `declinedChangeTopic` - Sent if user says no while in this broadcast topic
* `invalidChangeTopicResponse` - Sent if user replies with neither yes or no while in this broadcast topic
* `autoReply` - Auto reply sent after user has declined, and is now stuck in this topic (until they send a keyword or command to break out)

###  API changes

These types are available via the `GET /broadcasts` resource. A `GET /broadcasts` response has 2 new properties added:
 
* `type` - Corresponds to the broadcast content type

* `templates` - object property for when the broadcast is set to user's current topic. Empty for legacy  broadcasts.

This content type is a bit of a hybrid of both broadcast and topic, as it contains the outbound broadcast, but also contains replies for after the broadcast -- which is why it's being exposed as  a`GET /topics/:id`, and to get things working in Gambit Conversations by making the following changes:

* If the `broadcastId` sent to a `POST /messages?origin=broadcast` request is an `autoReplyBroadcast`, send the broadcast message and set the conversation.topic to the `broadcastId` sent.

* If a `POST /messages?origin=member` message is received for a user with an `autoReplyBroadcast` topic - send them the `autoReply` template.

As a sanity check while developing this, I thought a `GET /contentfulEntries/:id` endpoint would be useful - as it eliminates the guesswork of knowing which resource we're mapping a given ID to (and for an `autoReplyBroadcast`, it can show up as both ( which could be confusing 😕 ❓ I've taken a stab at updating the topics documentation accordingly).

I've created a sample `autoPostBroadcast` - 4nwTwvXmfuuYAGYgusGyyW, and will be opening up a Conversations PR to test against it.

* Verify that the `GET /broadcasts` response looks as expected for existing broadcasts, but also contains a `templates` object when type is set to `autoReplyBroadcast

 * Verify the `GET /contentfulEntries/:id` works as expected

It'd likely make sense to return these new content types in the `GET /topics` index response as well -- but it felt like more changes than necessary at the moment, and possibly confusing. Once we start adding many broadcast entries - our cached list of topics we use for sake of finding topics by campaign quickly would turn ugly. That code in generally is ugly and should be refactored to make separate Contentful calls to filter by campaign (either by adding custom query parameters, passing query params directly to Contentful, or potentially thinking GraphQL to DRY systems)

#### Relevant tickets
https://www.pivotaltracker.com/n/projects/2092729

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/config/lib/helpers/broadcast.js
+++ b/config/lib/helpers/broadcast.js
@@ -2,6 +2,7 @@
 
 module.exports = {
   broadcastContentTypes: [
+    'autoReplyBroadcast',
     'broadcast',
   ],
 };
--- a/config/lib/helpers/broadcast.js
+++ b/config/lib/helpers/broadcast.js
@@ -5,4 +5,5 @@ module.exports = {
     'autoReplyBroadcast',
     'broadcast',
   ],
+  legacyContentType: 'broadcast',
 };
--- a/config/lib/helpers/broadcast.js
+++ b/config/lib/helpers/broadcast.js
@@ -1,9 +1,24 @@
 'use strict';
 
 module.exports = {
-  contentTypes: [
-    'autoReplyBroadcast',
-    'broadcast',
-  ],
-  legacyContentType: 'broadcast',
+  contentTypes: {
+    askChangeTopicBroadcast: {
+      type: 'askChangeTopicBroadcast',
+      templates: [
+        'declinedChangeTopic',
+        'invalidAskChangeTopicResponse',
+        'autoReply',
+      ],
+    },
+    autoReplyBroadcast: {
+      type: 'autoReplyBroadcast',
+      templates: [
+        'autoReply',
+      ],
+    },
+    legacy: {
+      type: 'broadcast',
+      templates: [],
+    },
+  },
 };
--- a/config/lib/helpers/broadcast.js
+++ b/config/lib/helpers/broadcast.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-  broadcastContentTypes: [
+  contentTypes: [
     'autoReplyBroadcast',
     'broadcast',
   ],
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -74,7 +74,7 @@ function parseBroadcastInfoFromContentfulEntry(contentfulEntry) {
  */
 function parseBroadcastFromContentfulEntry(contentfulEntry) {
   const data = module.exports.parseBroadcastInfoFromContentfulEntry(contentfulEntry);
-  if (data.type === 'broadcast') {
+  if (data.type === config.legacyContentType) {
     return module.exports.parseLegacyBroadcastFromContentfulEntry(contentfulEntry);
   }
   return module.exports.parseAutoReplyBroadcastFromContentfulEntry(contentfulEntry);
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -19,12 +19,12 @@ function getContentTypes() {
  * @param {Object} query
  * @return {Promise}
  */
-function fetch(query = {}) {
-  return contentful.fetchByContentTypes(module.exports.getContentTypes(), query)
-    .then(contentfulRes => ({
-      meta: contentfulRes.meta,
-      data: contentfulRes.data.map(module.exports.parseBroadcastFromContentfulEntry),
-    }));
+async function fetch(query = {}) {
+  const contentfulRes = await contentful
+    .fetchByContentTypes(module.exports.getContentTypes(), query);
+  const broadcasts = await Promise
+    .all(contentfulRes.data.map(module.exports.parseBroadcastFromContentfulEntry));
+  return Object.assign(contentfulRes, { data: broadcasts });
 }
 
 /**
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -51,19 +51,42 @@ function getById(broadcastId) {
  * @param {Object} contentfulEntry
  * @return {Object}
  */
-function parseBroadcastFromContentfulEntry(contentfulEntry) {
-  const data = {
+function parseBroadcastInfoFromContentfulEntry(contentfulEntry) {
+  return {
     id: contentful.getContentfulIdFromContentfulEntry(contentfulEntry),
     name: contentful.getNameTextFromContentfulEntry(contentfulEntry),
+    type: contentful.getContentTypeFromContentfulEntry(contentfulEntry),
     createdAt: contentfulEntry.sys.createdAt,
     updatedAt: contentfulEntry.sys.updatedAt,
-    message: {
-      text: contentfulEntry.fields.message,
-      attachments: contentful.getAttachmentsFromContentfulEntry(contentfulEntry),
-    },
   };
-  // Note: We plan to split broadcast content type into two separate content types, so editors
-  // don't need to remember which fields to include/exclude in order to create a specific type.
+}
+
+/**
+ * @param {Object} contentfulEntry
+ * @return {Object}
+ */
+function parseBroadcastFromContentfulEntry(contentfulEntry) {
+  const data = module.exports.parseBroadcastInfoFromContentfulEntry(contentfulEntry);
+  if (data.type === 'broadcast') {
+    return module.exports.parseLegacyBroadcastFromContentfulEntry(contentfulEntry);
+  }
+  return module.exports.parseAutoReplyBroadcastFromContentfulEntry(contentfulEntry);
+}
+
+/**
+ * The 'broadcast' content type was used to send various types of broadcasts, which will soon be
+ * deprecated by new content types per broadcast type.
+ *
+ * @param {Object} contentfulEntry
+ * @return {Object}
+ */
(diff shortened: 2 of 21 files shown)


lib/middleware/topics/single/topic-get.js
```diff
@@ -3,7 +3,7 @@
 const helpers = require('../../../helpers');
 
 module.exports = function getTopic() {
-  return (req, res) => helpers.topic.getById(req.params.topicId)
+  return (req, res) => helpers.contentfulEntry.getById(req.params.topicId)
```
aaronschachter: This is a little bit of a hack, to avoid more changes than we need currently in the topic helper. Looking up an `askChangeTopicBroadcast` id would work here but we'd be missing the `message` property used for a broadcast, which seems confusing.

aaronschachter:
I've pushed up a commit that adds an additional  `askChangeTopicBroadcast` type of broadcast, and updated the PR description accordingly. This broadcast type hasn't been implemented in Conversations at all, so should not affect testing on https://github.com/DoSomething/gambit-conversations/pull/373. Going to push this branch to staging to test.

lib/helpers/contentfulEntry.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {Object} contentfulEntry
+ * @param {Promise}
+ */
+function parseContentfulEntry(contentfulEntry) {
+  // TODO: It might be nice to move all of these type of functions into this helper, to restrict
+  // lib/contentful to simply querying Contentful API.
+  const contentType = contentful.getContentTypeFromContentfulEntry(contentfulEntry);
+
+  if (helpers.broadcast.getContentTypes().includes(contentType)) {
+    return Promise.resolve(helpers.broadcast.parseBroadcastFromContentfulEntry(contentfulEntry));
```
rapala61: I think we should return with a Promise from `helpers.broadcast.parseBroadcastFromContentfulEntry` to keep these parsers consistent.

lib/helpers/broadcast.js
```diff
@@ -5,14 +5,21 @@ const contentful = require('../contentful');
...
+
 /**
  * Fetches broadcast Contentful entries and parses them as broadcasts.
  *
  * @param {Object} query
  * @return {Promise}
  */
 function fetch(query = {}) {
-  return contentful.fetchByContentTypes(config.broadcastContentTypes, query)
+  return contentful.fetchByContentTypes(module.exports.getContentTypes(), query)
```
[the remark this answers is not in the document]
aaronschachter: We always call functions with `module.exports` so they can easily be stubbed within unit tests.

lib/middleware/topics/single/topic-get.js
aaronschachter: I also thought about Conversations always querying the `GET /contentfulEntries/:id` instead of `/topics/:id`, but there's something nice and logical about Conversations querying the `GET /topics/:id` resource to find reply templates. It also had me wondering about renaming the `templates` property as  `replies` in our API response, e.g. 
```
    {
      "id": "4nwTwvXmfuuYAGYgusGyyW",
      "name": "Test autoReplyBroadcast",
      "type": "autoReplyBroadcast",
      "createdAt": "2018-07-25T14:26:03.277Z",
      "updatedAt": "2018-07-26T16:57:36.050Z",
      "message": {
        "text": "Hello, how are you today?",
        "attachments": [
        "template": "autoReplyBroadcast"
      },
      "replies": {
        "autoReply": {
          "text": "I'm doing well, thank you.",
          "attachments": [],
          "template": "autoReply"
        }
    }
 }
```

config/lib/helpers/broadcast.js
```diff
@@ -1,7 +1,24 @@
 'use strict';
 
 module.exports = {
-  broadcastContentTypes: [
-    'broadcast',
-  ],
+  contentTypes: {
+    askChangeTopicBroadcast: {
```
rapala61: I like how the content types are setup in the broadcast helper config. I think switching `topics` and `defaultTopicTriggers` to this scheme would simplify them (by adding more consistency) a bit more. Maybe is too much changes for this PR, but something to consider in future PRs.

lib/helpers/broadcast.js
```diff
@@ -80,9 +123,46 @@ function parseBroadcastFromContentfulEntry(contentfulEntry) {
   return data;
```
rapala61: We would also have to return with a Promise from this function so that both `legacy` and `new` broadcasts will return Promises from `parseBroadcastFromContentfulEntry`.

lib/middleware/topics/single/topic-get.js
rapala61: Minor. Can we add a `//TODO` comment on this hack if we think we are coming back and cleaning it up afterwards?

rapala61: Re: renaming the `templates` property as `replies`. I think renaming to `replyTemplates` could help clear up confusions about if by `template` we mean a content type template (as in the property inside the message object), or a reply template.

While we are in this piece of code. I would like to suggest calling that `template` property inside the `message` object `contentTypeTemplate` instead ? What do you think? 

Something like:
```
{
  "data": {
    "id": "4nwTwvXmfuuYAGYgusGyyW",
    "name": "Test autoReplyBroadcast",
    "type": "autoReplyBroadcast",
    "createdAt": "2018-07-25T14:26:03.277Z",
    "updatedAt": "2018-07-26T16:57:36.050Z",
    "message": {
      "text": "Hello, how are you today?",
      "attachments": [
        
      ],
      "contentTypeTemplate": "autoReplyBroadcast"
    },
    "replyTemplates": {
      "autoReply": {
        "text": "I'm doing well, thank you.",
        "attachments": [],
        "template": "autoReplyBroadcast"
      }
    }
  }
}
```

lib/helpers/broadcast.js
```diff
@@ -77,12 +120,49 @@ function parseBroadcastFromContentfulEntry(contentfulEntry) {
...
+}
+
+/**
+ * @param {Object} contentfulEntry
+ * @return {Object}
+ */
+function parseBroadcastMessageFromContentfulEntryAndTemplateName(contentfulEntry, templateName) {
+  const broadcastMessage = contentfulEntry.fields.broadcast;
+  return helpers.contentfulEntry
+    .getMessageTemplateFromContentfulEntryAndTemplateName(broadcastMessage, templateName);
```
rapala61: One more minor naming suggestion. Calling it `broadcastMessageEntry` or `broadcastEntry` could help imply that we are sending an actual entry object, without having previous knowledge of how these C.io schemas work. For a long time I though we were sending a string, until I consoled it out and realized it was a nested entry!

lib/middleware/topics/single/topic-get.js
aaronschachter: This `message.template` property here is what gets saved to the outbound message `template` property in Conversations... https://github.com/DoSomething/gambit-conversations/blob/2.14.1/lib/middleware/messages/broadcast/broadcast-parse.js#L10, and meant to match the objects contained in the `templates` property, as they are the reply messages we send back.
I'm starting to second guess renaming `templates` to `replies` or `replyTemplates`, as it'd make sense to change them on the `topics` response.  The gotcha there is that the `web*` templates aren't really replies, and are used for the outbound web signup messages. Similarly, I imagined the reminder messaging fields would be embedded in the `templates` well .. these would outbound and wouldn't make sense to live in `replies`.

lib/middleware/topics/single/topic-get.js
```diff
@@ -3,7 +3,9 @@
 const helpers = require('../../../helpers');
 
 module.exports = function getTopic() {
-  return (req, res) => helpers.topic.getById(req.params.topicId)
+  // TODO: Use helpers.topic.getById once the new broadcast content types have been added as topic
+  // content types. This is a hack to support saving new broadcasts as conversation topics.
+  return (req, res) => helpers.contentfulEntry.getById(req.params.topicId)
```
aaronschachter: This hack is 🚫 -- this `contenfulEntry` helper is not caching data. Working on a new branch to clean this up (and avoid the hack altogether)

aaronschachter: Fixed in #1064

PR #1064, opened 2018-07-27
Add broadcast topic types to topic content types [pr]

aaronschachter:
#### What's this PR do?
Cleanup from #1063: 

* Returns `autoReplyBroadcast` and `askChangeTopicBroadcast` types to the list of topic content types, so they are now returned as topics as well.

    * Undoes the hack that would've been bad news for a broadcast (uncached queries to GET /topics/:id) -- https://github.com/DoSomething/gambit-campaigns/pull/1063#discussion_r205825914

  * Begins to refactor the topic helper config per https://github.com/DoSomething/gambit-campaigns/pull/1063/files/769888be2e4834d06d859f4b15a77bab1a61c8d2#diff-13eba1df303e045dab5e41bbe9361751

   * This means the list of topics will grow after a week or two, which could complicate the creating cached list of topics. I'm pretty sure this exists so we can make a `GET /campaigns` request to find campaigns that currently have topics with triggers, which is only used in 2 places -- the MENU macro that looks to find something else for a user to do, and Gambit Admin. I'm inclined to look into something more manual to manage the MENU topics (e.g. add a field to the topic content types that collect posts whether it should be a menu option)

* Adds a `cache` query parameter to the `GET /broadcasts/:id` and `GET /topics/:id` endpoints to force a cache refresh

* Requires API key to be passed as header or query for all endpoints

   * Removes unused `v1/status` - I believe this was only added for a New Relic application availability test and has since been removed.

* Verify that `autoTopicBroadcast` and `autoReplyBroadcast` entries are returned in a `GET /topics` or `GET /topics/:id` response

* Verify expected behavior for passing (and not passing) a `cache=false` query to single topic and broadcast routes

* Verify all endpoints require an api key

I'll be opening up a Gambit Admin PR to authenticate requests. Curious about setting up a webhook from Contentful to ping the URLs accordingly, although it's slightly tricky that a broadcast entry can be cached in two places. I'm also considering adding a manual "clear cache" button into Gambit Admin based on level of effort of the Contentful webhook.

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158037562

#### Checklist
- [x] Tested on staging.


app/routes/index.js
```diff
@@ -32,7 +32,6 @@ function regGlobalMiddleware(app) {
 module.exports = function init(app) {
   regGlobalMiddleware(app);
   app.get('/', statusRoute);
-  app.use('/v1/status', statusRoute);
```
rapala61: We have a Runscope test that uses this route. https://www.runscope.com/radar/e6s8uo1kkh7r/bcd0fa4a-d5c2-492f-8b41-49f10f176ba4

lib/helpers/broadcast.js
```diff
@@ -35,23 +35,28 @@ async function fetch(query = {}) {
...
 
 /**
- * @param {String} topicId
+ * @param {String} broadcastId
+ * @param {Boolean} resetCache
  * @return {Promise}
  */
-function getById(broadcastId) {
+function getById(broadcastId, resetCache = false) {
+  if (resetCache) {
```
rapala61: Should also check strict equality here like we do with [topics](https://github.com/DoSomething/gambit-campaigns/pull/1064/files#diff-fe5513ac2c68bbe898ecb4288a671ac4R67)?

config/lib/helpers/topic.js
```diff
@@ -1,5 +1,33 @@
 'use strict';
 
+const contentTypes = {
```
rapala61: I think it would be a good idea to add some comments to explain why we are exposing broadcasts here as topics?

app/routes/index.js
aaronschachter: Great catch! I've updated the test.

PR #1065, opened 2018-08-06
Fixes bug in GET /broadcasts response for draft entries without a broadcast reference

aaronschachter:
#### What's this PR do?
Adds check for a `broadcast` field in the `parseBroadcastMessageFromContentfulEntryAndTemplateName` broadcast helper function, returning null if it doesn't exist. 

Create a draft `autoReplyBroadcast` entry without any fields. Verify that a `GET /broadcasts` query includes the entry, returning it with a null `message` property.

This field is required, so this error wouldn't happen on production -- but does require us to check for its existence to avoid 500 errors in QA.

#### Relevant tickets
Fixes https://dosomething.slack.com/archives/C1V0M6RPE/p1533575475000182

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -65,13 +65,15 @@ function getById(broadcastId, resetCache = false) {
  * @return {Object}
  */
 function parseBroadcastInfoFromContentfulEntry(contentfulEntry) {
-  return {
+  const result = {
     id: contentful.getContentfulIdFromContentfulEntry(contentfulEntry),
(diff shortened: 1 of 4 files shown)


aaronschachter:
Created entry 6zj3wladnGyWuMWAUYyg8c to test this against. Unfortunately Gambit Admin breaks, so need to verify via `GET /broadcasts` request passing an `apiKey` query or `x-gambit-api-key` header.

PR #1070, opened 2018-08-07
Renames askChangeTopicBroadcast to askYesNo

aaronschachter:
#### What's this PR do?

We're not using this in Conversations yet, so these are non-breaking changes to rename the content type and fields, and starts on coverage for the `contentfulEntry` helper.

*  It also checks for the existence of a `topic` reference field in `helpers.campaignActivity.getMessageTemplateFromContentfulEntryAndTemplateName` 
* Starts on moving over some helper functions out `lib/contentful` and into `lib/helpers/contentfulEntry`

Verify `GET /v1/broadcasts/2X4r3fZrTGA2mGemowgiEI` response contains templates specified in config, and that the `templates.saidYes.topic` property has an `id` property set.

Sure, I wrote up a [mini-spec in Google Docs](https://docs.google.com/document/d/1mKsIgYtheNsPbJK6hYjqUpUYkcmItnkjgxXb1GFKlOk/edit) to make it easier to discuss the fields for this `askYesNo` type,  as well as other additional content types.

#### Relevant tickets
https://www.pivotaltracker.com/story/show/157369418

#### Checklist
- [ ] Tested on staging.


lib/helpers/contentfulEntry.js
```diff
@@ -36,21 +37,59 @@ function parseContentfulEntry(contentfulEntry) {
...
+}
+
 /**
  * @param {Object} contentfulEntry
  * @param {String} templateName
  * @return {Object}
  */
 function getMessageTemplateFromContentfulEntryAndTemplateName(contentfulEntry, templateName) {
+  const topicRef = contentfulEntry.fields.topic;
+  const topic = topicRef ? { id: contentful.getContentfulIdFromContentfulEntry(topicRef) } : null;
```
rapala61: 2cents in this line: One thing I'm wondering is if sending an empty `{}` is more practical. This way other functions that expect properties to be set inside the `topic`, like `topic.id`, wouldn't have to first check for existence ` result.topic ? topic.id : null`. Also probably helps with unwanted `Cannot read property 'id' of null`.

aaronschachter: Good call! Will update

PR #1099, opened 2018-11-16
Removes deprecated topics array on GET /campaigns/:id response

aaronschachter:
#### What's this PR do?

Now that Conversations no longer queries a `GET /campaigns/:id` response to find the first topic within a `topics` array response, we can remove the `topics` array property completely -- which was brittle to begin with since it attempted to cache a page of results (but only stopped at a single page)

Upon staging deploy, nothing should change. Conversations doesn't ever query for the `topics` array on a `GET /campaigns/:id` response -- this can be tested manually by participating in a campaign post conversation (e.g. text in LOCK or MIRROR)... getting topic to change to `random` via entering topics like `askSubscriptionStatus` (e.g. text `status`), and then verifying expected behavior per https://github.com/DoSomething/gambit-conversations/pull/441 deployment that a `noCampaign` response is returned upon answering with a valid subscription status.

* The plan moving forward for finding topics by a campaignId (which would mainly take place in the Gambit Admin repo) would be to query the `GET /contentfulEntries` endpoint by the relevant Contentful types and field values.

* We're in good shape to deprecate the `GET /topics` and `GET /campaigns` index endpoints entirely, but holding off on that in hopes of keeping this PR easier to review. 

#### Checklist
- [x] Tested on staging.


DFurnes:
> The plan moving forward for finding topics by a campaignId (which would mainly take place in the Gambit Admin repo) would be to query the `GET /contentfulEntries` endpoint by the relevant Contentful types and field values.

This makes good sense to me. 👌

PR #1101, opened 2018-11-30
Templates cleanup

aaronschachter:
#### What's this PR do?

Removes the templates that we no longer use anymore, which will allow us to delete these fields in Contentful once this goes live:

* `photoPostConfig.startPhotoPost`, `textPostConfig.askText`- These templates are now only sent by via broadcast or default topic trigger transition

* All `campaign` template fields:

    * `campaignClosed` - Deprecated in https://github.com/DoSomething/gambit-conversations/pull/388

    * `askContinue`, `declinedContinue`, `invalidAskContinueResponse` - Deprecated in https://github.com/DoSomething/gambit-conversations/pull/441

Also completes TODO of DRY the topic and contentfulEntry helpers that map a Contentful entry's field to the `templates` object includes in responses of our broadcast and topic API endpoints.

* Modifies the contentfulEntry `getTopicTemplates` helper to provide a hardcoded value for optional photo post fields, like `askWhyParticipated`. Removes topic `getTopicTemplates` helper to DRY.

* **Note**: This removes the `override` and `rendered` properties, so for now, we'd need to open an entry in Contentful to see whether a handful of optional photo post templates are hardcoded defaults.

* Verify expected responses in  `GET /broadcasts/:id`,  `GET /topics/:id` requests

Part 2: Cleanup the `helpers.replacePhoenixVars` function, possibly remove `helpers/command` and build support into Conversations for a `commands` tag, e.g. `Text {{comands.start}} when you're ready to submit`

#### Checklist
- [x] Tested on staging.


lib/helpers/contentfulEntry.js
```diff
@@ -138,10 +138,13 @@ async function getTopicTemplates(contentfulEntry) {
 
   const templateNames = Object.keys(templatesConfig);
   const promises = templateNames.map((templateName) => {
```
rapala61: Minor observation. Using [Promise.map](http://bluebirdjs.com/docs/api/promise.map.html#promise.map) here could condense this code a bit more.

PR #1107, opened 2019-01-18
Adds faqAnswer type to deprecate message type

aaronschachter:
#### What's this PR do?

I've added a new Gambit content type called `faqAnswer`, to replace what we use the current `message` content type for -- answering frequently asked questions like "Are you a bot" or "Who is this"  via `defaultTopicTrigger` entries.

There are only a handful of these entries, so I'll manually migrate each one over as a `faqAnswer` once this PR is merged to production.

This `message` type is from way before the days of transition types and is long overdue for a name change (#1094).

Here's an example `faqAnswer` entry I set up for "How old are you?":

screen shot 2019-01-18 at 3 12 33 pm

 Once this branch is published, our `what is your age [*]` trigger can reference this instead of the `message` entry it currently references (`4ePqvl7koUo6a2qeSCO0kQ`)

I've set a `testing` keyword to draft so we can test on Gambit Staging. 

screen shot 2019-01-18 at 3 18 40 pm

**Send `testing` and verify the "Age is just a number" reply is sent.**

screen shot 2019-01-18 at 3 22 11 pm

In getting ready to move `defaultTopicTrigger` entries into GraphQL as `conversationTriggers` -- it'd be simpler to avoid redirects to other `defaultTopicTrigger` entries. This means that the `response` reference field on a `defaultTopicTrigger` would only be able to reference the following types:
* `autoReplyTransition`
* `photoPostTransition`
* `textPostTransition`
* `faqAnswer`

#### Relevant tickets
https://github.com/DoSomething/graphql/pull/42

#### Checklist

- [x] Tested on staging.

### DIFF
--- a/lib/helpers/contentfulEntry.js
+++ b/lib/helpers/contentfulEntry.js
@@ -223,8 +223,9 @@ function isLegacyBroadcast(contentfulEntry) {
  * @param {Object} contentfulEntry
  * @return {Boolean}
  */
-function isMessage(contentfulEntry) {
-  return contentful.isContentType(contentfulEntry, config.contentTypes.message.type);
+function isFaqAnswer(contentfulEntry) {
+  return [config.contentTypes.message.type, config.contentTypes.faqAnswer.type]
+    .includes(contentful.getContentTypeFromContentfulEntry(contentfulEntry));
 }
 
 /**
@@ -260,7 +261,7 @@ module.exports = {
   isBroadcastable,
   isDefaultTopicTrigger,
   isLegacyBroadcast,
-  isMessage,
+  isFaqAnswer,
   isTransitionable,
   isTransitionTemplate,
   parseContentfulEntry,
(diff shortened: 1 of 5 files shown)


lib/helpers/contentfulEntry.js
```diff
@@ -223,8 +223,9 @@ function isLegacyBroadcast(contentfulEntry) {
  * @param {Object} contentfulEntry
  * @return {Boolean}
  */
-function isMessage(contentfulEntry) {
-  return contentful.isContentType(contentfulEntry, config.contentTypes.message.type);
+function isFaqAnswer(contentfulEntry) {
+  return [config.contentTypes.message.type, config.contentTypes.faqAnswer.type]
```
rapala61: Just a nit, it would be a bit more readable this way:
```
  return [
    config.contentTypes.message.type,
    config.contentTypes.faqAnswer.type,
    ]
```

rapala61 (approved):

lib/helpers/contentfulEntry.js
aaronschachter: Agreed -- I'm going to leave for now as this code should hopefully be toast soon 🍞 . It's been great having `prettier` integrated over in our [https://github.com/DoSomething/graphql](GraphQL) linting to catch and fix things like these... definitely would be worth adding into Conversations!

